PR #61, opened 2017-07-06
Fix #58: Authorization notification using WebSockets vs. regular authentication in mtoken step

### LINKED ISSUE #58: Authorization notification using WebSockets vs. regular authentication in mtoken step
Currently we use a 3s polling using a call of the authenticate() method. At the same time a notification from WebSockets can arrive causing another call of the authenticate() method. We should handle the case when the time of both operations would coincide - avoid calling the authenticate() method twice.

romanstrobl:
Potential race condition in the authenticate() method calls is handled

I tested both WebSocket authorization and 3s polling, both work fine.

### DIFF
--- a/powerauth-webauth/src/main/js/components/token.js
+++ b/powerauth-webauth/src/main/js/components/token.js
@@ -40,11 +40,11 @@ export default class Token extends React.Component {
         this.init = this.init.bind(this);
         this.onRegister = this.onRegister.bind(this);
         this.onAuthorize = this.onAuthorize.bind(this);
-        this.setAuthorizationByWebSocketInProgress = this.setAuthorizationByWebSocketInProgress.bind(this);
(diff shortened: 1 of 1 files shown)


romanstrobl:
Hm, finally after fixing the potential issue I was able to see it manifest:

2017-07-08 22:49:54.840  WARN 4479 --- [nio-8080-exec-4] o.h.engine.jdbc.spi.SqlExceptionHelper   : SQL Error: 1062, SQLState: 23000
2017-07-08 22:49:54.841 ERROR 4479 --- [nio-8080-exec-4] o.h.engine.jdbc.spi.SqlExceptionHelper   : Duplicate entry 'e3cc32a8-b9d2-4055-ad1f-843f02b56956-5' for key 'PRIMARY'

romanstrobl:
I fixed the locking, it should handle the case when authorization is already in progress and websocket message arrives.

PR #64, opened 2017-07-08
Fix #29: Separate different FAILED states

### LINKED ISSUE #29: Separate different FAILED states
We need to separate the FAILED state for an authentication method from a completely failed method. 

New states will be:
* AUTH_FAILED
* AUTH_METHOD_FAILED

We will mock AUTH_METHOD_FAILED state in the Credential store after 5 failed attempts to login with the same username.

### LINKED ISSUE #55: Implement central check for failed authorization attempts
Currently, we rely on individual methods to count number of failures and canceling the operation.

We should add support for failed operations to Next Step as well...

romanstrobl:
Note that the DDL and initial data has changed and it is incompatible with previous data because of change of AuthResult: FAILED->AUTH_FAILED (to separate AUTH_FAILED and AUTH_METHOD_FAILED states).

Furthemore, the number of allowed authorization attempts and whether to check failures at all can be configured in table ns_auth_method (columns check_auth_fails and max_auth_fails).

Also note that FAILED states are separted in AuthStepResult, AuthResult didn't change.

Full changelog:
* Fix #29: Separate different FAILED states, as described above.
* Fix #55: Implement central check for failed authorization attempts - each method in NS can be configured with the number of allowed authorization attempts, if it is exceeded the method fails.
* Updated schema and DDL for new AuthStepResult and new transitions in NS - it is better to drop tables and load data again.
* Support for maximum number of authentication attempts during login - this showcases the NS support for checking number of failed attempts.
* Configuration of maximum number of FAILED results per method saved in database, as described above.
* Partial implementation of fallbacks - in case two authentication methods are available and one fails, the non-failing authentication method is preferred. We may need to improve the UI to handle this.
* Failed attempts are now logged into operation history when buildAuthorizationResponse() method is called and authenticate() method returns null. We don't do this for the mtoken step, partial failures are just polling.
* Method failAuthorization() can be called from auth methods to log failed attempt. This is shown in FormLoginController.


romanstrobl:
Here is screenshot of operation history when login operation fails with AUTH_METHOD_FAILED after maximum number of attempts is exceeded.

screen shot 2017-07-08 at 21 32 20

PR #70, opened 2017-07-17
Fix #36: Implement support for invalid SSL certificate on Web Flow related services

### LINKED ISSUE #36: Implement support for invalid SSL certificate on Web Flow related services
We currently expect communication with either valid SSL certificate or over HTTP.

Web Flow application should support `powerauth.service.ssl.acceptInvalidSslCertificate` attribute for communication with PowerAuth 2.0 Server and similar attributes for other application it communicates with.

romanstrobl:
Added trusting of all certificates during initializations of various clients


powerauth-webauth-authentication/src/main/java/io/getlime/security/powerauth/lib/webauth/authentication/configuration/WebAuthServicesConfiguration.java
```diff
@@ -57,7 +81,44 @@ public CredentialStoreClient defaultCredentialStoreClient() {
...
+        if (acceptInvalidSslCertificateInNextStep) {
+            trustAllCertificates();
+        }
+        return client;
     }
 
+    /**
+     * Activate trust in all SSL certificates including invalid ones for non-production use.
+     */
+    private void trustAllCertificates() {
```
petrdvorak: This method disables SSL certificate handling everywhere - not just on one or another interface. We should either:

- implement SSL disabling in such way that it works per client (Next Step vs. Credential Store), or...
- have only one property "acceptInvalidSslCertificate" that disables SSL check for all services.

powerauth-webauth-client/src/main/java/io/getlime/security/powerauth/app/webauth/demo/configuration/WebAuthServiceConfiguration.java
```diff
@@ -73,7 +86,45 @@ public String getClientSecret() {
...
+            trustAllCertificates();
+        }
+        return client;
+    }
+
+    /**
+     * Activate trust in all SSL certificates including invalid ones for non-production use.
+     */
+    private void trustAllCertificates() {
+        HttpsURLConnection.setDefaultHostnameVerifier((hostname, session) -> true);
```
petrdvorak: Same issue as above - this setting works for all `HttpsURLConnection` instances, there is no benefit in having two separate attributes for it and current behavior is surprising ("I only disabled SSL on Next Step, and it's disabled on Credential Store as well), hence incorrect.

powerauth-webauth/src/main/java/io/getlime/security/powerauth/app/webauth/configuration/PowerAuthWebServiceConfiguration.java
```diff
@@ -60,14 +70,55 @@ public PowerAuthServiceClient powerAuthClient(Jaxb2Marshaller marshaller) {
...
+        if (acceptInvalidSslCertificate) {
+            trustAllCertificates();
+        }
         return client;
     }
 
+    /**
+     * Activate trust in all SSL certificates including invalid ones for non-production use.
+     */
+    private void trustAllCertificates() {
```
petrdvorak: Why do we disable SSL in multiple Maven modules in a single project?

powerauth-webauth-authentication/src/main/java/io/getlime/security/powerauth/lib/webauth/authentication/configuration/WebAuthServicesConfiguration.java
romanstrobl: You are right. I think it is enough to have one property, this setting is used for development as far as I understand.

powerauth-webauth/src/main/java/io/getlime/security/powerauth/app/webauth/configuration/PowerAuthWebServiceConfiguration.java
romanstrobl: It is not the same project - one commit affects powerauth-webauth.war and second commit affects powerauth-webauth-client.war. I thought we needed to disable SSL certificate check in each war file separately? If not, then I will change it.

romanstrobl: Ah, I see now, you probably mean duplicate setting in powerauth-webauth and powerauth-webauth-authentication projects. Yes, that should be fixed.

powerauth-webauth-client/src/main/java/io/getlime/security/powerauth/app/webauth/demo/configuration/WebAuthServiceConfiguration.java
romanstrobl: Fixed, same configuration property is used in all projects.

powerauth-webauth-authentication/src/main/java/io/getlime/security/powerauth/lib/webauth/authentication/configuration/WebAuthServicesConfiguration.java
```diff
@@ -43,12 +47,28 @@
...
+    /**
      * Default credential store client.
      * @return Credential store client.
      */
     @Bean
     public CredentialStoreClient defaultCredentialStoreClient() {
-        return new CredentialStoreClient(credentialServerServiceUrl);
+        CredentialStoreClient client = new CredentialStoreClient(credentialServerServiceUrl);
+        // whether invalid SSL certificates should be accepted
+        if (acceptInvalidSslCertificate) {
```
petrdvorak: We are not disabling SSL validation for the client, but for all server application...

PR #74, opened 2017-07-18
Fix #68: Allow custom operation ID while creating operation

### LINKED ISSUE #68: Allow custom operation ID while creating operation
In case a system calling the Next Step service would like to provide other than the default operation ID when creating operation, we probably should not restrict it. For example, API gateway may generate unique "API call ID" for each requests, and this API call ID may need to enter into unified logging.

In case client application provides operation ID while creating operation in Next Step server, we should:

- check if such operation ID exists and...
    - if yes, we should reject the operation creation
    - if no, we should use the operation ID and create the new operation

romanstrobl:
Note that this pull request contains pull requests #69 and #71 because of potential merge conflicts.

It should be possible to use a custom operationId now when creating an operation.

### DIFF
--- a/powerauth-nextstep-client/src/main/java/io/getlime/security/powerauth/lib/nextstep/client/NextStepClient.java
+++ b/powerauth-nextstep-client/src/main/java/io/getlime/security/powerauth/lib/nextstep/client/NextStepClient.java
@@ -91,28 +91,31 @@ public class NextStepClient {
      * Calls the operation endpoint via POST method to create a new operation.
      *
      * @param operationName operation name
+     * @param operationId operation ID (optional)
      * @param operationData operation data
      * @param params        list of generic parameters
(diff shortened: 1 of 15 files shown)


powerauth-nextstep-client/src/main/java/io/getlime/security/powerauth/lib/nextstep/client/NextStepClient.java
```diff
@@ -91,28 +91,31 @@ private RestTemplate defaultTemplate() {
      * Calls the operation endpoint via POST method to create a new operation.
      *
      * @param operationName operation name
+     * @param operationId operation ID (optional)
```
petrdvorak: Lazy fix. :) If the parameter is optional, create a method that does not have it - as an API consumer, how do I know what "(optional)" means here?

PR #91, opened 2017-09-21
Refactoring of project name: Web Auth -> Web Flow 

romanstrobl:
Refactoring includes packages, class names, properties, Maven module names, etc.
Fixed formatting issues from IntelliJ IDEA (around @ResponseBody annotation)


romanstrobl:
Unfortunately IntelliJ IDEA reformatted some parts of the code, althrough I already unchecked formatting during commit, sorry about that. I reverted the @ResponseBody formatting changes at least, the rest seems to be ok.

romanstrobl:
Just for the record - there is a problem with existing OAuth2 access tokens which are serialized into database. The serialized classes contain the 'webauth' package name. It is necessary to clean the database to avoid compability problems for the client application.

PR #119, opened 2017-10-11
Fix #118: Make the demo app more independent

### LINKED ISSUE #118: Make the demo app more independent
The demo app can have in-memory database and does not have to run in Tomcat server.

petrdvorak:
Use in-memory H2 DB
Allow launching via 'java -jar' command

### DIFF
--- a/powerauth-webflow-client/pom.xml
+++ b/powerauth-webflow-client/pom.xml
@@ -84,6 +84,13 @@
             <version>0.15.0</version>
         </dependency>
 
+        <!-- H2 DB -->
+        <dependency>
+            <groupId>org.hsqldb</groupId>
+            <artifactId>hsqldb</artifactId>
+            <scope>runtime</scope>
+        </dependency>
+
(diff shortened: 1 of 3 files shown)


romanstrobl:
I get an exception when I start the app using

$ java -jar powerauth-webflow-client-0.15.0.war

o.s.boot.SpringApplication               : Application startup failed

org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'entityManagerFactory' defined in class path resource [org/springframework/boot/autoconfigure/orm/jpa/HibernateJpaAutoConfiguration.class]: Invocation of init method failed; nested exception is java.lang.NoClassDefFoundError: javax/xml/bind/JAXBException

Perhaps this is about classpath setting?

petrdvorak:
Works fine on the lime server and my localhost, probably some issue with cleaning the workspace?

romanstrobl:
Maybe, I will try it later. Other than this the change looks good.

Issue #98, opened 2017-10-01
Push message sent even to user who does not have mToken active


petrdvorak:
We currently sent message based on userId, without checking if the push message should be actually sent. As a result, user receives push message without having mToken enabled.

romanstrobl:
Issue should be resolved by https://github.com/lime-company/powerauth-webflow/issues/120, which will not show Mobile Token button in case there are no activations available for the user.

PR #124, opened 2017-10-19
Fix #120: Add check for actual authentication method availability

### LINKED ISSUE #120: Add check for actual authentication method availability
Currently, we only check user preferences to determine availability of an authentication method. For example, if user activates mobile token, we show a mobile token UI despite the fact user has no active devices.

We should add a generic `isAvailable` check to authentication method base, that would use user preferences as one of the inputs, but also check the actual method availability based on user and operation data...

```java
boolean isAvailable(String userId, String operationId);
```

romanstrobl:
This pull request fixes issues #120 and #98 related to dynamic change of availability of authentication methods e.g. when activations get blocked.

I also added error handling for initialization of authentication methods and fixed minor issues such as javadoc and i18n messages.


romanstrobl:
Please wait with the review, I want to clean up the error handling.

romanstrobl:
The pull request is ready for review now.

powerauth-webflow-authentication-operation-review/src/main/java/io/getlime/security/powerauth/lib/webflow/authentication/method/operation/controller/OperationReviewController.java
```diff
@@ -61,14 +60,21 @@
...
     private final String FIELD_CHOSEN_AUTH_METHOD = "chosenAuthMethod";
 
-    @Autowired
-    private DataAdapterClient dataAdapterClient;
+    private final DataAdapterClient dataAdapterClient;
+    private final NextStepClient nextStepClient;
 
-    @Autowired
-    private NextStepClient nextStepClient;
+    public OperationReviewController(DataAdapterClient dataAdapterClient, NextStepClient nextStepClient) {
```
petrdvorak: Missing `@Autowired` above the constructor. If it works, it is by an accident.

romanstrobl:
I added the missing @Autowired annotation.

PR #126, opened 2017-10-20
Fix #107: Isolate business logic of Data Adapter into separate service

### LINKED ISSUE #107: Isolate business logic of Data Adapter into separate service
We currently have business logic spread across multiple controllers. We should introduce an interface, let customers implement it according to their business logic, and autowire the service in controllers.

romanstrobl:
Business logic is isolated. We can discuss how to further separate implementation code for the 3rd party, for now I placed client specific code into the 'impl' package.

### DIFF
--- a/powerauth-data-adapter/src/main/java/io/getlime/security/powerauth/lib/dataadapter/impl/service/DataAdapterService.java
+++ b/powerauth-data-adapter/src/main/java/io/getlime/security/powerauth/lib/dataadapter/impl/service/DataAdapterService.java
@@ -1,0 +1,272 @@
+package io.getlime.security.powerauth.lib.dataadapter.impl.service;
+
+import io.getlime.security.powerauth.crypto.server.util.DataDigest;
+import io.getlime.security.powerauth.lib.dataadapter.api.DataAdapter;
+import io.getlime.security.powerauth.lib.dataadapter.configuration.DataAdapterConfiguration;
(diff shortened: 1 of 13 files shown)


powerauth-data-adapter/src/main/java/io/getlime/security/powerauth/lib/dataadapter/impl/service/DataAdapterService.java
```diff
@@ -0,0 +1,272 @@
...
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * Sample implementation of DataAdapter interface which should be updated in real implementation.
+ *
+ * @author Roman Strobl
+ */
+@Service
+public class DataAdapterService implements DataAdapter {
```
petrdvorak: Not nice. You have basically just moved the code from one class to another - this does not make much sense. The idea was to extract the **custom logic** into separate service, not the whole code. Simply put, the code that should be extracted should only implement the logic that is required for customization. For example, DataAdapterService must not contain SMSAuthorizationRepository, because the customizing client knows nothing about how it works. Instead, he should have a method:

`boolean sendSms(user, message)`

... that just does the job using the stuff he know how works. Etc.

Since the issue is related mainly to the SMS sending, I will probably accept this pull request and then we will figure out how to improve the code.

PR #128, opened 2017-10-23
Fix #107: Isolate business logic of Data Adapter into separate service

### LINKED ISSUE #107: Isolate business logic of Data Adapter into separate service

romanstrobl:
Reimplemented with a better interface for consumers.

### DIFF
--- a/powerauth-data-adapter/src/main/java/io/getlime/security/powerauth/lib/dataadapter/api/DataAdapter.java
+++ b/powerauth-data-adapter/src/main/java/io/getlime/security/powerauth/lib/dataadapter/api/DataAdapter.java
@@ -39,7 +39,7 @@ public interface DataAdapter {
      * @param password Password for user authentication.
      * @throws AuthenticationFailedException Thrown when authentication fails.
      */
-    void authenticateUser(String username, String password) throws AuthenticationFailedException;
+    UserDetailResponse authenticateUser(String username, String password) throws AuthenticationFailedException;
 
     /**
      * Fetch user detail for given user.
--- a/powerauth-data-adapter/src/main/java/io/getlime/security/powerauth/lib/dataadapter/api/DataAdapter.java
+++ b/powerauth-data-adapter/src/main/java/io/getlime/security/powerauth/lib/dataadapter/api/DataAdapter.java
@@ -17,15 +17,16 @@ package io.getlime.security.powerauth.lib.dataadapter.api;
 
 import io.getlime.security.powerauth.lib.dataadapter.exception.AuthenticationFailedException;
 import io.getlime.security.powerauth.lib.dataadapter.exception.SMSAuthorizationFailedException;
-import io.getlime.security.powerauth.lib.dataadapter.model.request.*;
-import io.getlime.security.powerauth.lib.dataadapter.model.response.AuthenticationResponse;
-import io.getlime.security.powerauth.lib.dataadapter.model.response.BankAccountListResponse;
-import io.getlime.security.powerauth.lib.dataadapter.model.response.CreateSMSAuthorizationResponse;
+import io.getlime.security.powerauth.lib.dataadapter.exception.UserNotFoundException;
+import io.getlime.security.powerauth.lib.dataadapter.model.entity.BankAccount;
+import io.getlime.security.powerauth.lib.dataadapter.model.entity.FormDataChange;
+import io.getlime.security.powerauth.lib.dataadapter.model.entity.OperationChange;
 import io.getlime.security.powerauth.lib.dataadapter.model.response.UserDetailResponse;
-import org.springframework.web.bind.MethodArgumentNotValidException;
+
+import java.util.List;
 
 /**
- * DataAdapter interface defines methods which should be implemented for integration of Web Flow with 3rd parties.
+ * Interface defines methods which should be implemented for integration of Web Flow with 3rd parties.
  *
  * @author Roman Strobl, roman.strobl@lime-company.eu
  */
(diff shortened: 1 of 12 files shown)


powerauth-data-adapter/src/main/java/io/getlime/security/powerauth/lib/dataadapter/api/DataAdapter.java
```diff
@@ -34,55 +35,50 @@
...
      *
-     * @param authenticationRequest Authentication request.
-     * @return Authentication response.
-     * @throws MethodArgumentNotValidException Thrown when input validation fails.
+     * @param username Username for user authentication.
+     * @param password Password for user authentication.
      * @throws AuthenticationFailedException Thrown when authentication fails.
      */
-    AuthenticationResponse authenticateUser(AuthenticationRequest authenticationRequest) throws MethodArgumentNotValidException, AuthenticationFailedException;
+    void authenticateUser(String username, String password) throws AuthenticationFailedException;
```
petrdvorak: We should return something from this method, for example an object with user information (at least, user ID has to be returned). It feels weird that we only let the method go and let it fail on authentication failure...

powerauth-data-adapter/src/main/java/io/getlime/security/powerauth/lib/dataadapter/controller/AuthenticationController.java
```diff
@@ -58,7 +63,18 @@ public AuthenticationController(DataAdapter dataAdapter) {
...
+        AuthenticationRequestValidator validator = new AuthenticationRequestValidator();
+        BeanPropertyBindingResult result = new BeanPropertyBindingResult(authenticationRequest, "authenticationRequest");
+        ValidationUtils.invokeValidator(validator, authenticationRequest, result);
+        if (result.hasErrors()) {
+            // getEnclosingMethod() on new object returns a reference to current method
+            MethodParameter methodParam = new MethodParameter(new Object(){}.getClass().getEnclosingMethod(),0);
+            throw new MethodArgumentNotValidException(methodParam, result);
+        }
+        dataAdapter.authenticateUser(authenticationRequest.getUsername(), authenticationRequest.getPassword());
+        AuthenticationResponse response = new AuthenticationResponse(authenticationRequest.getUsername());
```
petrdvorak: This is the issue - "username" is not "userId". Generally, these values differ. While username in a typical bank is either a client number, e-mail or username, it always translates to canonical user ID...

powerauth-data-adapter/src/main/java/io/getlime/security/powerauth/lib/dataadapter/controller/BankAccountController.java
```diff
@@ -51,9 +52,9 @@ public BankAccountController(DataAdapter dataAdapter) {
      * @return Response with user details.
      */
     @RequestMapping(value = "/list", method = RequestMethod.POST)
-    public @ResponseBody ObjectResponse<BankAccountListResponse> fetchBankAccounts(@RequestBody ObjectRequest<BankAccountListRequest> request) throws MethodArgumentNotValidException {
+    public @ResponseBody ObjectResponse<BankAccountListResponse> fetchBankAccounts(@RequestBody ObjectRequest<BankAccountListRequest> request) throws MethodArgumentNotValidException, UserNotFoundException {
         BankAccountListRequest bankAccountListRequest = request.getRequestObject();
-        BankAccountListResponse response = dataAdapter.fetchBankAccounts(bankAccountListRequest);
+        BankAccountListResponse response = new BankAccountListResponse(bankAccountListRequest.getUserId(), dataAdapter.fetchBankAccounts(bankAccountListRequest.getUserId()));
```
petrdvorak: This code could use a little polist... for example:

```java
String userId = bankAccountListRequest.getUserId();
// ...
```

Generally, I prefer slightly more lines of code with a clear meaning, than large method parameter chaining...

powerauth-data-adapter/src/main/java/io/getlime/security/powerauth/lib/dataadapter/controller/SMSAuthorizationController.java
```diff
@@ -54,8 +64,21 @@ public SMSAuthorizationController(DataAdapter dataAdapter) {
      */
     @RequestMapping(value = "create", method = RequestMethod.POST)
     public @ResponseBody ObjectResponse<CreateSMSAuthorizationResponse> create(@RequestBody ObjectRequest<CreateSMSAuthorizationRequest> request) throws MethodArgumentNotValidException, SMSAuthorizationFailedException {
-        CreateSMSAuthorizationRequest createSMSAuthorizationRequest = request.getRequestObject();
-        CreateSMSAuthorizationResponse response = dataAdapter.createAuthorizationSMS(createSMSAuthorizationRequest);
+        CreateSMSAuthorizationRequest smsRequest = request.getRequestObject();
+        // input validation is handled by CreateSMSAuthorizationRequestValidator
+        // validation is invoked manually because of the generified Request object
+        BeanPropertyBindingResult result = new BeanPropertyBindingResult(smsRequest, "createSMSAuthorizationRequest");
```
petrdvorak: I do not understand this code - do we really need to use custom binding?

powerauth-data-adapter/src/main/java/io/getlime/security/powerauth/lib/dataadapter/controller/SMSAuthorizationController.java
```diff
@@ -54,8 +64,21 @@ public SMSAuthorizationController(DataAdapter dataAdapter) {
...
+        ValidationUtils.invokeValidator(requestValidator, smsRequest, result);
+        if (result.hasErrors()) {
+            // getEnclosingMethod() on new object returns a reference to current method
+            MethodParameter methodParam = new MethodParameter(new Object() {
```
petrdvorak: This also seems like something we should not need in the code.

powerauth-data-adapter/src/main/java/io/getlime/security/powerauth/lib/dataadapter/impl/service/DataAdapterService.java
```diff
@@ -39,47 +22,23 @@
...
-            MethodParameter methodParam = new MethodParameter(new Object(){}.getClass().getEnclosingMethod(),0);
-            throw new MethodArgumentNotValidException(methodParam, result);
-        }
-
-        // here will be the real authentication - call to the backend providing authentication
-        if ("test".equals(authenticationRequest.getPassword())) {
-            return new AuthenticationResponse(authenticationRequest.getUsername());
-        } else {
-            throw new AuthenticationFailedException("login.authenticationFailed");
+    public void authenticateUser(String username, String password) throws AuthenticationFailedException {
```
petrdvorak: dtto as above ...

romanstrobl:
I have resolved the comments about return value from user authentication call and avoiding calling methods in method arguments.

The validation code is a bit more tricky. I replaced manual Validator invocation by injecting specific Validators into controllers and doing validation on the ObjectRequests. Luckily Spring supports field nesting and it works with generified ObjectRequest, so this code works to my surprise:

```
ValidationUtils.rejectIfEmptyOrWhitespace(errors, "requestObject.username", "login.username.empty");
```

The error handling for failed validation is still a bit cumbersome (throwing the MethodArgumentNotValidException), we can look for a cleaner solution, however current code should work.

PR #141, opened 2017-11-01
Fix #131: Change user/info endpoint

### LINKED ISSUE #131: Change user/info endpoint
Based on the comments, we should update the response to:

```json
{
    "user": {
        "id": "1234567890",
        "givenName": "John",
        "familyName": "Doe"
   },
    "connection": {
        "language": "CZ",
        "sca":true
    }
}
```

- `user.*` - the user ID, name and surname as we have it
- `connection.language` - language used for issuing the OAuth 2.0 access token
- `connection.sca` - true, if the OAuth 2.0 access token was issued using strong customer authentication, false otherwise

petrdvorak:


petrdvorak:
Fixed.

PR #156, opened 2017-11-03
Fix #153: Get application ID from PowerAuthApiAuthentication instead of configuration

### LINKED ISSUE #153: Get application ID from PowerAuthApiAuthentication instead of configuration
Currently applicationId is configured in application.properties for push registration.

We can extract applicationId from PowerAuthApiAuthentication object instead.

romanstrobl:


powerauth-webflow-authentication-mtoken/src/main/java/io/getlime/security/powerauth/lib/webflow/authentication/mtoken/controller/MobileTokenOnlineController.java
```diff
@@ -125,8 +131,37 @@ protected AuthMethod getAuthMethodName() {
...
+            return initResponse;
+        }
+
+        Long applicationId = null;
+
+        // loading of activations
+        List<GetActivationListForUserResponse.Activations> allActivations = powerAuthServiceClient.getActivationListForUser(userId);
+
+        // resolve applicationId from activationId
+        for (GetActivationListForUserResponse.Activations activation: allActivations) {
```
petrdvorak: This can be simplified by using `getActivationStatus` method.

romanstrobl:
I swtiched to the getActivationStatus() method both in online and offline mode.

PR #174, opened 2017-11-05
#172 - Form Data Attribute ids and i18n

### LINKED ISSUE #171: Account picker disabled after switching language
Steps to reproduce:

1. Create a new payment operation
2. Log in using username and password
=> You see payment overview
=> Account picker is enabled

3. Click the language choser
=> Account picker is disabled

Expected result: Account picker remains enabled after switching the language

### LINKED ISSUE #172: Label for account does not use the i18n key in extraction for SMS
The label is currently defined as follows when it is extracted for SMS authorization:

LABEL_TO_ACCOUNT = "To Account";

Due to localization the label should be changed to: "operation.account".

romanstrobl:
Changes:

- Fix #172: Label for account does not use the i18n key in extraction for SMS
- OperationFormAttributes have ids now which are used both for i18n and for unique identification of attributes
- Created OperationFormAttributes for title and message for easier localization (operation message is called MAIN_MESSAGE to separate it from payment MESSAGE - we should revise the MESSAGE types together with Mobile app, for compatibility title and message fields are still present in OperationFormData, we can remove this redundancy, too)
- Message translation code uses attribute ids (the only problematic value remains currency which needs its own extra id in OperationAmountAttribute for identification of currency field in SMS messages, but I guess this is ok)
- Added several utility methods into OperationFormData class
- Client code simplified thanks to OperationFormData utility methods and attribute ids
- Fix #171: Account picker disabled after switching language

There are some TODOs related to i18n which I cannot fix until we also fix them in mobile app, because the fields are used. We can do it together later this week.


powerauth-nextstep-model/src/main/java/io/getlime/security/powerauth/lib/nextstep/model/entity/OperationFormData.java
```diff
@@ -12,133 +15,246 @@
...
-     * @param label Attribute label.
+     * @param id Attribute ID.
      * @param value Attribute value.
      */
-    public void addKeyValue(String label, String value) {
+    @JsonIgnore
+    public void addKeyValue(String id, String value) {
         OperationKeyValueAttribute attr = new OperationKeyValueAttribute();
-        attr.setLabel(label);
+        attr.setId(id);
```
petrdvorak: I don't get it - we still need "label", in order to be able to localize the label, right? We cannot do it by ID, since it is a lookup / addressing identifier that has nothing to do with localization...

powerauth-nextstep-model/src/main/java/io/getlime/security/powerauth/lib/nextstep/model/entity/attribute/OperationFormAttribute.java
```diff
@@ -0,0 +1,99 @@
...
+        KEY_VALUE
+    }
+
+    // JsonIgnore added, otherwise type was serialized twice
+    @JsonIgnore
+    protected Type type;
+
+    protected String id;
+
+    protected String label;
```
petrdvorak: "label" does not feel like generic attribute key, it is related to the particular display of the attribute of given type... "type" and "id" should be enough here.

powerauth-nextstep-model/src/main/java/io/getlime/security/powerauth/lib/nextstep/model/entity/attribute/OperationFormAttribute.java
```diff
@@ -0,0 +1,99 @@
...
+    public enum Type {
+        TITLE,
+        MAIN_MESSAGE,
+        AMOUNT,
+        MESSAGE,
+        BANK_ACCOUNT_CHOICE,
+        KEY_VALUE
+    }
+
+    // JsonIgnore added, otherwise type was serialized twice
```
petrdvorak: This seems a bit suspicious...

powerauth-webflow/src/main/js/components/operationDetail.js
```diff
@@ -103,7 +103,7 @@ export default class OperationDetail extends React.Component {
                         {this.props.context.formData.parameters.map((item) => {
                             if (item.type === "AMOUNT") {
                                 return (
-                                    <div className="row attribute" key={item.label}>
+                                    <div className="row attribute" key={item.id}>
```
petrdvorak: Label and ID have completely different meaning. There must be no two items with the same ID. But there can be many items with the same label...

romanstrobl: Exactly, which is why we should use ID in this place, because react requires they key to be unique.

petrdvorak: Sorry... I see it now.

powerauth-nextstep-model/src/main/java/io/getlime/security/powerauth/lib/nextstep/model/entity/attribute/OperationFormAttribute.java
romanstrobl: Introduction of generic label made the internationalization much easier - I can just call attribute.setLabel(localize(id)) but I can put it back into individual subclasses, it doesn't matter so much.

powerauth-nextstep-model/src/main/java/io/getlime/security/powerauth/lib/nextstep/model/entity/attribute/OperationFormAttribute.java
romanstrobl: If @JsonIgnore is missing, type is present twice. Perhaps it is related to @JsonTypeInfo? I can investigate it.

powerauth-nextstep-model/src/main/java/io/getlime/security/powerauth/lib/nextstep/model/entity/attribute/OperationFormAttribute.java
petrdvorak: OK, it makes sense. Thank you for explanation.

powerauth-nextstep-model/src/main/java/io/getlime/security/powerauth/lib/nextstep/model/entity/OperationFormData.java
romanstrobl: As discussed below, I made label a generic field. Then by calling:

String localizedValue = localize(id);
attribute.setLabel(localizedValue);

we can simply localize all OperationFormAttributes.

But if you think it does not make sense to have labels for some OperationFormAttributes, method setLabel() can be called depending on actual availability of the label.

petrdvorak: OK - true. Your way is better.

powerauth-nextstep-model/src/main/java/io/getlime/security/powerauth/lib/nextstep/model/entity/attribute/OperationFormAttribute.java
petrdvorak: Yes. Let's have a look on it...

PR #186, opened 2017-11-11
Fix #41: Test API calls of auth methods for availability in current step

### LINKED ISSUE #41: Test API calls of auth methods for availability in current step
For example, if someone calls `/auth/mtoken/operation/detail` endpoint before the login (via username / password), we currently do not do any check that would stop the call.

romanstrobl:
Fix #41: Test API calls of auth methods for availability in current step
- Logic for validating operation state is now centralized including error handling
- Additional operation state checks both in Web Flow and Next Step
- Method getOperation() no longer returns null, AuthStepException is thrown instead
- Special handling for operation continuity checks for SHOW_OPERATION_DETAIL auth method


powerauth-nextstep/src/main/java/io/getlime/security/powerauth/app/nextstep/service/StepResolutionService.java
```diff
@@ -359,6 +359,28 @@ private void checkLegitimityOfUpdate(OperationEntity operationEntity, UpdateOper
         if (initOperationItem.getRequestAuthMethod() != null || initOperationItem.getRequestAuthStepResult() != null) {
             throw new IllegalStateException("Operation update failed, because INIT step for this operation is invalid (operationId: " + request.getOperationId() + ").");
         }
+        // check whether request AuthMethod is available in response AuthSteps - this verifies operation continuity
+        boolean stepAuthMethodValid = false;
+        if (request.getAuthMethod() == AuthMethod.SHOW_OPERATION_DETAIL) {
+            // special handling for SHOW_OPERATION_DETAIL - either SMS_KEY or POWERAUTH_TOKEN are present in next steps
+            for (AuthStep step: operationPersistenceService.getResponseAuthSteps(operationEntity)) {
+                if (step.getAuthMethod() == AuthMethod.SMS_KEY || step.getAuthMethod()==AuthMethod.POWERAUTH_TOKEN) {
```
petrdvorak: ... inconsistent formatting... ;-) Use spaces around `==`...

PR #196, opened 2017-11-16
Fix #194: Simplify input of offline token and add format validations

### LINKED ISSUE #194: Simplify input of offline token and add format validations
In order to simplify entry of the offline token values, let's format the input field so that user enters 4x4 characters, into "4 fake textfields", just as you might know it from the card number entry components.

The code will be then translated into "XXXXXXXX-XXXXXXXX" format on the server, so that the format matches 2FA signature.

Also, we need to validate the format before we call server validation methods and in case format is invalid so that we don't let 2MB text file (or any other issue) in our database...

romanstrobl:


romanstrobl:
The pull request is not ready yet - the offline code input component needs to handle properly key and mouse actions.

romanstrobl:
All cases I could test work properly now. You can try to break formatting inside the component.

PR #207, opened 2017-11-24
Fix #54: Handle multiple concurrent operations

### LINKED ISSUE #54: Handle multiple concurrent operations
It should not be possible to start two concurrent operations. In case a second operation is started, the first one should be canceled.

Also, handle two opened tabs in the web UI.

romanstrobl:
Older operation is canceled when new operation is initiated within same HTTP session. The reaction in older tab occurs when next action is perfomed by the user, in Mobile Token step the cancellation is visible almost immediately because of polling. We can improve the user experience in future by introducing Web Sockets, but I guess this is quite a niche scenario that user starts two operations in same browser.


powerauth-nextstep-client/src/main/java/io/getlime/security/powerauth/lib/nextstep/client/NextStepClient.java
```diff
@@ -96,8 +96,8 @@ private RestTemplate defaultTemplate() {
      * @param params        list of generic parameters
      * @return a Response with CreateOperationResponse object for OK status or ErrorModel for ERROR status
      */
-    public ObjectResponse<CreateOperationResponse> createOperation(String operationName, String operationId, String operationData, List<KeyValueParameter> params) throws NextStepServiceException {
-        return createOperation(operationName, operationId, operationData, null, params);
+    public ObjectResponse<CreateOperationResponse> createOperation(String operationName, String operationId, String operationData, String httpSessionId, List<KeyValueParameter> params) throws NextStepServiceException {
```
petrdvorak: This will not work. How does system outside the browser know about HTTP session? Method `createOperation` is called from outside the web browser most of the time (payment is created via microservice adapter running at bank back-end)...

petrdvorak: (In other words: HTTP session must be checked somewhere on webflow level.)

powerauth-webflow-authentication/src/main/java/io/getlime/security/powerauth/lib/webflow/authentication/controller/AuthMethodController.java
```diff
@@ -77,17 +80,23 @@
...
+
     protected GetOperationDetailResponse getOperation() throws AuthStepException {
         final UserOperationAuthentication pendingUserAuthentication = authenticationManagementService.getPendingUserAuthentication();
         if (pendingUserAuthentication != null) {
             String operationId = pendingUserAuthentication.getOperationId();
             if (operationId != null) {
                 return getOperation(operationId);
             } else {
-                return null;
+                throw new AuthStepException("operation.notAvailable", new NullPointerException());
```
petrdvorak: This looks weird - do not use `new NullPointerException()`, just make sure to log the warning and provide appropriate `message` for the exception. Currently, we are incorrectly "hacking" `message` to pass a localization key - this should be used for providing readable information in the exception.  We can either introduce new exception subclass `class OperationNotAvailableException extends AuthStepException`, or extend `AuthStepException` by adding `code` field with extended info about error that caused this exception to be raised.

powerauth-webflow-authentication/src/main/java/io/getlime/security/powerauth/lib/webflow/authentication/repository/model/entity/OperationSessionEntity.java
```diff
@@ -0,0 +1,109 @@
...
+
+/**
+ * Entity which stores mapping of operations to HTTP sessions and operation results.
+ *
+ * @author Roman Strobl, roman.strobl@lime-company.eu
+ */
+@Entity
+@Table(name = "wf_operation_session")
+@NamedQueries({
+        @NamedQuery(name = "OperationSessionEntity.findActiveOperationsByHttpSessionId", query = "SELECT o FROM OperationSessionEntity o WHERE o.httpSessionId = ?1 AND o.result='CONTINUE'")
```
petrdvorak: Rather question than a comment - what do we get by using NamedQuery atop of the Entity, not inside the Repository implementation?

powerauth-webflow-authentication/src/main/java/io/getlime/security/powerauth/lib/webflow/authentication/service/OperationSessionService.java
```diff
@@ -0,0 +1,94 @@
+package io.getlime.security.powerauth.lib.webflow.authentication.service;
```
petrdvorak: Missing copyright header.

powerauth-webflow-authentication/src/main/java/io/getlime/security/powerauth/lib/webflow/authentication/controller/AuthMethodController.java
petrdvorak: ... we have it more times... I will open separate issue on this.

romanstrobl: You are right, we should do the error handling more properly, this is a misuse of the AuthStepException class.

powerauth-webflow-authentication/src/main/java/io/getlime/security/powerauth/lib/webflow/authentication/repository/model/entity/OperationSessionEntity.java
romanstrobl: I've seen it in articles and tutorials done like this. I don't feel strongly about this issue, we can move it to Repository in next release. It is actually quite nice to see the NamedQuery right next to the method in Repository, so I like your suggestion.

powerauth-webflow-authentication/src/main/java/io/getlime/security/powerauth/lib/webflow/authentication/service/OperationSessionService.java
romanstrobl: Good catch, fixed.

romanstrobl:
One more thing we discussed is in which project should be the OperationSessionEntity class stored. Right now all JPA-related classes for handling concurrency are in the powerauth-webflow-authentication project, which is the base project for all authentication methods. We can move the classes directly into the powerauth-webflow-authentication-init project, they are quite related to the INIT auth method, but this would mean some small refactoring of the operation initialization code. We also discussed moving the model class out of the "lib" project to the "app" project, however this would mean adding a dependency, so I do not think it is worth it.

I suggest we leave the JPA classes as they are (from my point of view the current placement doesn't cause any problems) and reconsider this issue in next release once we add more DB tables, entities, repositories, etc.

Issue #219, opened 2017-11-30
Isolate network objects from business logic objects


petrdvorak:
Our localization "mumbo jumbo" is to large extent caused by the fact that we incorrectly reuse objects between various software layers. In ideal world, we have (simplified):

```java
// Business level

class OperationBO {
   private String title;
}

OperationBO operationBO = new OperationBO();
operationBO.setTitle("payment.localization.key")

/// ...

// Network level
class OperationVO {
    private String title;
}
OperationVO operationVO = new OperationVO();
operationVO.setTitle(i18nService.getKey(operationBO.getTitle(), locale));
```

romanstrobl:
I agree, but we should do this change with future OperationFormData redesign in mind. See also related issue:
https://github.com/lime-company/powerauth-webflow/issues/221

petrdvorak:
This did not prove to be a relevant enhancement. Closing for now, feel free to reopen if the situation changes.

PR #220, opened 2017-11-30
Fix #218: Make the default login operation more detailed

### LINKED ISSUE #218: Make the default login operation more detailed
Currently, we default to "login" operation with "{}" data in case no operation ID is provided.

We should add some default form data as well, so that in case this operation is configured to use 2FA, we have something to display on the mobile device.

petrdvorak:


petrdvorak:
... I am almost sure this implementation has issues, due to the way we localize form data. I believe what we need is to use "operation.title" as ID for the form title attribute, but we need to set text to something else than "Confirm Payment".

I have opened #219, that is a root cause...

powerauth-webflow-authentication-init/src/main/java/io/getlime/security/powerauth/lib/webflow/authentication/method/init/controller/ApiController.java
```diff
@@ -58,8 +71,12 @@
         if (operation == null) {
             final String operationName = "login";
             final String operationData = "{}";
+            AbstractMessageSource messageSource = i18nService.getMessageSource();
+            final OperationFormData formData = new OperationFormData();
+            formData.addTitle("operation.title", "login.title");
+            formData.addMessage("operation.message", "login.message");
```
romanstrobl: In current implementation you can use the following code:
formData.addTitle("login.title");
formData.addMessage("login.message");

This way you set the ID of i18n key for different title and message depending on the operation. So if you want to initialize payment instead of login operation, you can use:
formData.addTitle("payment.title");
formData.addMessage("payment.message");

And so on.

As long as the i18n keys are present, this will work, because Web Flow will localize the value as well as translate it using the {xyz} patterns.

I realize that this implementation is not ideal, because we want to separate ID for i18n from ID of the form field to make title and message work a bit more like web input fields. However currently the JS frontend supports only one title and one message (and mToken app as well), so the IDs of these form fields are quite irrelevant. I suggest to postpone extending OperationFormData to future release.

I raised issue to extend OperationFormData and separate ID of field from ID for i18n:
https://github.com/lime-company/powerauth-webflow/issues/221

petrdvorak:
OK, can I merge this one? :)

romanstrobl:
Yes

PR #222, opened 2017-11-30
Fix #208: Implement possession signed endpoints with using token based authentication

### LINKED ISSUE #208: Implement possession signed endpoints with using token based authentication.
Recently, we have successfully implemented and tested a token based authentication in our mobile client application, so we should remove that old possesion-signed endpoints and replace them with token based authentication.
* `/api/auth/token/app/push/register` -> currently has token impl. at `/api/auth/token/app/token/push/register`
* `/api/auth/token/app/operation/list` -> currently has token impl. at `/api/auth/token/app/token/operation/list`

As you can see, there are too much "token" in the URL, so we should just use the old URLs with this new token base method for authentication.

romanstrobl:
Both new token endpoints replace the old non-token endpoints.

### DIFF
--- a/powerauth-webflow-authentication-mtoken/src/main/java/io/getlime/security/powerauth/lib/webflow/authentication/mtoken/controller/MobileAppApiController.java
+++ b/powerauth-webflow-authentication-mtoken/src/main/java/io/getlime/security/powerauth/lib/webflow/authentication/mtoken/controller/MobileAppApiController.java
@@ -65,20 +65,6 @@ public class MobileAppApiController extends AuthMethodController<MobileTokenAuth
     }
 
     /**
-     * Register device for push notifications using 1FA signature.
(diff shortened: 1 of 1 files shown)


petrdvorak:
I will wait for a while with merging these - we have customer playing with the original endpoints and signature based authentication (not in production) and we need to synchronize the information with the customer...

petrdvorak:
OK, another brief discussion about these changes.

- We currently have a customer waiting for the push service registration to be implemented.
- This customer will not be able to update to (still non-released) mobile libraries 0.17.0 in time.

Therefore, we need to keep both endpoints and we just need to clean up the naming a bit...

petrdvorak:
Suggestion: Let's move push registration related endpoints into separate controller, so that we can have some nice URI, like `/api/push/device/register/token` and `/api/push/device/register/signature`...

PR #240, opened 2017-12-05
Fix #213: Make sure each application has "healthcheck" service

### LINKED ISSUE #213: Make sure each application has "healthcheck" service
Every application we publish must have a healthcheck service, so that bank is able to see the service is up.

See PA Push Server service for example:

https://github.com/lime-company/powerauth-push-server/wiki/Push-Server-API#service
https://github.com/lime-company/powerauth-push-server/blob/develop/powerauth-push-server/src/main/java/io/getlime/push/controller/rest/ServiceController.java#L51

romanstrobl:


powerauth-webflow/src/main/resources/application.properties
```diff
@@ -43,4 +43,9 @@ spring.jpa.properties.hibernate.connection.useUnicode=true
 #spring.jpa.properties.hibernate.temp.use_jdbc_metadata_defaults=false
 
 # Hibernate Configuration
-spring.jpa.hibernate.ddl-auto=create
\ No newline at end of file
+spring.jpa.hibernate.ddl-auto=create
```
petrdvorak: ... not related to this change, but we probably should place "none" here. The reason we are not randomly deleting application data is that we override this variable in Docker image setup.

PR #264, opened 2017-12-18
Fix #262: Add version and compilation time to service status endpoint

### LINKED ISSUE #262: Add version and compilation time to service status endpoint

romanstrobl:
This is a quick addition of the version and compilation time. I would like to postpone the refactoring of placing service status API code into a library after the release. Right now there are few lines of duplicate code, but I don't think this is a problem now.


romanstrobl:
Yes, no need to reinvent the wheel.

PR #281, opened 2018-01-17
Fix #279: Allow default attributes for dynamic form data

### LINKED ISSUE #279: Allow default attributes for dynamic form data
Currently, we show the dynamic form data (account picker) anytime account list is fetched for given operation in data adapter. However, we do not allow some basic form features, such as:

- disabling the dynamic picker
- selecting a default value

I suggest extending our form API for creating a new operation in a following way:

```json
{
  "formData": {
    "attributes": [
      {
        "id": "operation.bankAccountChoice",
        "enabled": false,
        "defaultValue": "CZ0012341234123412341234"
      }
    ]
  }
}
```

This would indicate, that form field with ID `operation.bankAccountChoice` should be disabled and should display item that corresponds to the default value `CZ0012341234123412341234`.

_Design choice: Keeping these attributes separate from the form fields can be a subject of discussion - I placed them this way so that the dynamic data can be customized as well._

romanstrobl:
Added possibility to disable choice and set default value for bank account select component.

By default choice is enabled and default value is not set, the UI chooses the first item in case default value is not set or it is invalid.


powerauth-webflow-authentication-operation-review/src/main/java/io/getlime/security/powerauth/lib/webflow/authentication/method/operation/controller/OperationReviewController.java
```diff
@@ -261,16 +262,16 @@ private OperationFormData loadFormData(GetOperationDetailResponse operation) {
...
-     * @param bankAccountEntities List of BankAccount entities.
+     * @param bankAccountList List of BankAccount entities.
      * @return List of BankAccountDetail.
      */
-    private List<BankAccountDetail> convertBankAccountEntities(List<BankAccount> bankAccountEntities) {
+    private List<BankAccountDetail> convertBankAccountList(BankAccountList bankAccountList) {
         // TODO - move to a converter class
         List<BankAccountDetail> bankAccountDetails = new ArrayList<>();
-        if (bankAccountEntities==null || bankAccountEntities.isEmpty()) {
+        if (bankAccountList == null || bankAccountList.getBankAccounts().isEmpty()) {
```
petrdvorak: Possible NPE on `bankAccountList.getBankAccount() == null`.

powerauth-webflow-authentication-operation-review/src/main/java/io/getlime/security/powerauth/lib/webflow/authentication/method/operation/controller/OperationReviewController.java
```diff
@@ -261,16 +262,16 @@ private OperationFormData loadFormData(GetOperationDetailResponse operation) {
...
             return bankAccountDetails;
         }
-        for (BankAccount bankAccountEntity: bankAccountEntities) {
+        for (BankAccount bankAccountEntity: bankAccountList.getBankAccounts()) {
```
petrdvorak: Possible NPE.

PR #292, opened 2018-02-07
Fix #287: Wrong text in QR code in mobile API

### LINKED ISSUE #287: Wrong text in QR code in mobile API
We should separate message into two fields in operation form data:
* greeting - Hello,\n please confirm following payment:
* summary - Hello, please confirm payment {operation.amount} {operation.currency} to account {operation.account}.

Currently message field is used for both use cases and thus we have incorrect summary text in push message and in the QR code.

romanstrobl:


romanstrobl:
Good idea, I will prepare the migration page.

romanstrobl:
See:
https://github.com/lime-company/powerauth-webflow/wiki/Migration-Instructions

PR #294, opened 2018-02-08
Fix #94: Make Web Flow compilable on JDK 9

### LINKED ISSUE #94: Make Web Flow compilable on JDK 9
Currently Web Flow does not compile on the newly released JKD 9.

We should look into it.

romanstrobl:


romanstrobl:
Current status: Web Flow builds on JDK 9 and can be deployed to Tomcat on Java 9. Tomcat needs to contain following libraries in lib folder (available from Maven Central). These libraries are former part of Java which are no longer visible in Java 9.

istack-commons-runtime-3.0.5.jar
javax.activation-api-1.2.0.jar
javax.xml.soap-api-1.4.0.jar
jaxb-api-2.3.0.jar
jaxb-core-2.3.0.jar
jaxb-impl-2.3.0.jar
jaxws-api-2.3.0.jar
saaj-api-1.3.5.jar
saaj-impl-1.4.0.jar
txw2-2.3.0.jar

All of these libraries can be downloaded from here:
[java9_javaee_libs.zip](https://github.com/lime-company/powerauth-webflow/files/1711519/java9_javaee_libs.zip)

Web Flow runs without problems on Java 9, however we PowerAuth doesn't support Java 9 yet due to problematic Bouncy Castle registration in Java 9 for multiple applications using Bouncy Castle running on one Tomcat.

PR #303, opened 2018-02-28
Fix #302: Invalid error handling during OAuth 2.0 handshake

### LINKED ISSUE #302: Invalid error handling during OAuth 2.0 handshake
Web Flow should handle "error" and "error_description" query parameters added by Spring security when OAuth 2.0 handshake fails:
https://github.com/spring-projects/spring-security-oauth/blob/master/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/endpoint/AuthorizationEndpoint.java#L367

The redirect logic in Web Flow is handled here:
https://github.com/lime-company/powerauth-webflow/blob/develop/powerauth-webflow/src/main/java/io/getlime/security/powerauth/app/webflow/controller/HomeController.java#L167

We should handle the error based on:
https://www.oauth.com/oauth2-servers/authorization/the-authorization-response

We should also clear the context when the error occurs.

romanstrobl:
Added error and error_description query parameters when operation is canceled by user.

### DIFF
--- a/powerauth-webflow/src/main/java/io/getlime/security/powerauth/app/webflow/controller/HomeController.java
+++ b/powerauth-webflow/src/main/java/io/getlime/security/powerauth/app/webflow/controller/HomeController.java
@@ -31,6 +31,8 @@ import org.springframework.web.bind.annotation.RequestMethod;
 
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.util.Locale;
 import java.util.Map;
 import java.util.logging.Level;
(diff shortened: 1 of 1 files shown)


powerauth-webflow/src/main/java/io/getlime/security/powerauth/app/webflow/controller/HomeController.java
```diff
@@ -181,7 +185,24 @@ public String cancelAuthentication(HttpServletRequest request, HttpServletRespon
             authenticationManagementService.clearContext();
         }
 
-        return "redirect:" + redirectUri;
+        try {
```
petrdvorak: This is too manual - use `UriComponentsBuilder` instead.

https://docs.spring.io/autorepo/docs/spring-framework/4.3.5.RELEASE/javadoc-api/org/springframework/web/util/UriComponentsBuilder.html

You can use it like this:

```java
final String fixedCallbackUrl = UriComponentsBuilder.fromUriString(callbackUrl)
                            .queryParam("redirectId", redirectId).build()
                            .toUriString();
```

romanstrobl:
This PR is still shown as open, can we close it?

romanstrobl:
Nevermind, previous comment closed the PR.

PR #305, opened 2018-03-02
Fix #245: Date and amount formatting by locale

### LINKED ISSUE #245: Date and amount formatting by locale
We currently localize form keys. For some values, however, we should allow value localization as well. These include date and amount with currency.

Technically, we should introduce new form items for these value types.

(soon post release)

romanstrobl:


powerauth-nextstep-model/src/main/java/io/getlime/security/powerauth/lib/nextstep/model/entity/attribute/OperationFormFieldAttributeFormatted.java
```diff
@@ -0,0 +1,50 @@
+package io.getlime.security.powerauth.lib.nextstep.model.entity.attribute;
```
petrdvorak: Missing copyright notice.

powerauth-webflow-authentication/src/main/java/io/getlime/security/powerauth/lib/webflow/authentication/service/ValueFormatterService.java
```diff
@@ -0,0 +1,209 @@
+package io.getlime.security.powerauth.lib.webflow.authentication.service;
```
petrdvorak: Missing copyright notice.

PR #323, opened 2018-03-27
Prepare release 0.19.0

petrdvorak:


petrdvorak:
Here are the basic release notes to summarize changes:

- Add custom attribute value formatting (date, amount, localized string, etc.).
- Separate model module for Mobile Token services.
- Fixed issue with incorrect OAuth 2.0 error state redirection.
- JDK9 support.
- Greeting and operation summary instead of a single generic operation message.
- Moved named queries from entities to repositories to clean up code.
- Improved and cleaned up exception handling.
- Bug fixes and dependency updates.
- JavaDoc improvements.

... it is a good release. ;-)

romanstrobl:
One more change which has impact on customization:

- CSS is split into two files: `base.css` and `customization.css` to simplify CSS customization

PR #332, opened 2018-04-06
Fix #318: Support for warnings in operation form data

### LINKED ISSUE #304: Add OperationContext to all methods in Data Adapter interface
OperationContext should contain operation details such as:
* operationId
* operationName
* operationData
* operationFormData

### LINKED ISSUE #318: Support for warnings in operation form data
Currently, we only allow operation to proceed or to fail. However, it could be useful to have some intermediate warning states, for example for cases when source account is not correct and hence combobox with choice contains the first item.

We should implement two types of warnings:

- "banner message" - a message with icon and outline displayed above the operation data form, with color and message. See [examples on Bootstrap website](https://getbootstrap.com/docs/4.0/components/alerts/#examples).
    - technically, we should add `banners` collection to form data, where each banner has type and message
- "in-form message" - a message with icon and outline displayed in-line inside the form, for example above the field it relates to
     - technically, the `BANNER_INFO`, `BANNER_WARNING` and `BANNER_ERROR` attribute types (or some equivalent mechanism) could be added to form while being decorated, see lime-company/powerauth-webflow-customization#8 for details.

romanstrobl:
Note that this PR does not try to fix the DA model or move DA project. I will do this in separate PR, because this PR is already quite big.

Additional changes:
* Fix #304: Add OperationContext to all methods in Data Adapter interface
* Replaced fetchBankAccounts method by decorateFormData method (issue #8 in powerauth-webflow-customization)
* Minor fixes 
   * localization of unexpected error messages
   * updated validators
   * simplified form attribute localization
   * updated JavaDoc
   * ValueFormatType enumeration moved out of attribute class


powerauth-data-adapter-model/src/main/java/io/getlime/security/powerauth/lib/dataadapter/model/request/DecorateOperationFormDataRequest.java
```diff
@@ -14,34 +14,34 @@
...
 
-package io.getlime.security.powerauth.lib.dataadapter.model.response;
+package io.getlime.security.powerauth.lib.dataadapter.model.request;
 
-import io.getlime.security.powerauth.lib.dataadapter.model.entity.BankAccountList;
+import io.getlime.security.powerauth.lib.dataadapter.model.entity.OperationContext;
 
 /**
- * Response with list of user bank accounts.
+ * Request object for obtaining bank accounts by user ID.
```
petrdvorak: I think this comment is no longer accurate, since this is now a generic form data decoration request.

PR #346, opened 2018-04-18
Fix #344: Add "heading" attribute support

### LINKED ISSUE #344: Add "heading" attribute support
We should add heading support into operation form data.

See:
https://github.com/lime-company/mtoken-ios/issues/110

romanstrobl:


powerauth-webflow/src/main/js/components/operationDetail.js
```diff
@@ -173,6 +173,14 @@ export default class OperationDetail extends React.Component {
                                         </div>
                                     </div>
                                 )
+                            } else if (item.type === "HEADING") {
+                                return (
+                                    <div className="row attribute" key={item.id}>
+                                        <div className="col-xs-6 key">
```
petrdvorak: Use `col-xs-12` class - the heading can then span over full row width.

PR #382, opened 2018-06-26
Fix #381: Enable 1FA signatures on operation verification endpoint

### LINKED ISSUE #381: Enable 1FA signatures on operation verification endpoint

petrdvorak:


petrdvorak:
... wait a second with the review, I will rewrite the core PR method so that the implicit result is `false`...

petrdvorak:
Done.

PR #449, opened 2018-11-08
Fix #448: Deprecate push server encryption

### LINKED ISSUE #448: Deprecate push server encryption
The `PushMessageAttributes.setEncrypted(boolean)` method was removed, we should remove it from WebFlow.

romanstrobl:


petrdvorak:
Looks OK

PR #476, opened 2019-02-14
Fix #420: Limit "amount" values to 2 decimal places

### LINKED ISSUE #420: Limit "amount" values to 2 decimal places
It's possible to send amount with value that has more than 2 decimal places in Amount Attribute in operation. This behavior should be sanitized on the server.

romanstrobl:
Added validation for amount attribute.

### DIFF
--- a/powerauth-nextstep-model/src/main/java/io/getlime/security/powerauth/lib/nextstep/model/entity/validator/AmountValidator.java
+++ b/powerauth-nextstep-model/src/main/java/io/getlime/security/powerauth/lib/nextstep/model/entity/validator/AmountValidator.java
@@ -1,0 +1,31 @@
+package io.getlime.security.powerauth.lib.nextstep.model.entity.validator;
+
+import io.getlime.security.powerauth.lib.nextstep.model.exception.InvalidOperationDataException;
+
+import java.math.BigDecimal;
+
+/**
+ * Validator for amount attribute.
+ *
+ * @author Roman Strobl, roman.strobl@wultra.com
+ */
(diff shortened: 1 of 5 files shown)


powerauth-nextstep-model/src/main/java/io/getlime/security/powerauth/lib/nextstep/model/entity/validator/AmountValidator.java
```diff
@@ -0,0 +1,31 @@
...
+     */
+    public static void validateAmount(BigDecimal amount) throws InvalidOperationDataException {
+        if (amount == null) {
+            throw new InvalidOperationDataException("Amount was not set");
+        }
+        if (amount.compareTo(BigDecimal.ZERO) <= 0) {
+            throw new InvalidOperationDataException("Amount is zero or negative: " + amount);
+        }
+        // Reject amount with more than 2 decimal places.
+        if (amount.stripTrailingZeros().scale() > 2) {
```
petrdvorak: I would make this a parameter of the validation method so that some operations can work with better precision.

```java
public static void validateAmount(BigDecimal amount) {
    return validateAmount(amount, 2);
} 
public static void validateAmount(BigDecimal amount, int expectedPrecision) throws InvalidOperationDataException {
    // ...
}
```

romanstrobl: Ok, sure.

PR #483, opened 2019-02-26
Fix #462: Display a security warning on Android devices

### LINKED ISSUE #462: Display a security warning on mobile devices
In case the user opens the login page on mobile, we should display a warning banner with security information to the user. This banner should be displayed at the very top of the page, and is not related to the operation / user flow context.

romanstrobl:
This is a draft pull request to finalize the mobile warning screen.

TODO list:
- ~~Decide whether we should display warning only on Android or all mobile devices (`isAndroid` constant can be easily replaced by `isMobile`, see all options [here](https://github.com/duskload/react-device-detect/blob/master/index.d.ts)).~~
  - The warning will be displayed only on Android devices.
- ~~Decide whether the warning override should be permanent - currently it is displayed in beginning of each operation.~~
  - We will display the warning in the beginning of each new operation.
- Update CSS and texts.


romanstrobl:
I attach HTML sources so that the CSS and texts can be updated easily without need to deploy the app from issue branch.


[override.zip](https://github.com/wultra/powerauth-webflow/files/2910135/override.zip)

romanstrobl:
We are starting to get conflicts in these long living branches. I will start merge these pull requests and we can focus on the UI as an improvement.

PR #505, opened 2019-05-10
Fix #504: Improve docs of off-line signatures

### LINKED ISSUE #504: Improve docs of off-line sigatures

petrdvorak:


docs/Off-line-Signatures-QR-Code.md
```diff
@@ -192,6 +211,55 @@ Generate the QR code from `offlineData`. Code example in Java:
...
+        BaseEncoding.base64().decode(nonce),
+        data.getBytes()
+);
+```
+
+#### 3.4. Verifying Signature
+
+To verify signature, you need to call the SOAP method [`verifyOfflineSignature`](https://developers.wultra.com/docs/develop/powerauth-server/SOAP-Service-Methods#method-verifyofflinesignature) providing:
+
+- `signatureBaseString` (as obtained in 3.2.) - as data for verification
```
romanstrobl: It would be better to match the parameter order from `verifyOfflineSignature` endpoint, the first parameter is `activationId` and so on...

docs/Off-line-Signatures-QR-Code.md
```diff
@@ -192,6 +211,55 @@ Generate the QR code from `offlineData`. Code example in Java:
             return "data:image/png;base64," + BaseEncoding.base64().encode(bytes);
 ```
 
-### 5. Verify ECDSA signature of offlineData
+The user can now scan the QR code via the mobile token app.
+
+The value of `nonce` must be stored somewhere - for example on a browser level, either as a hidden HTML form input field or as a React.js variable, or alternatively, it can be stored in the user session by the specific operation. The `nonce` value is required for later offline signature validation, see step 3.
```
romanstrobl: Not sure we need to mention React.js, using `JavaScript variable` is more generic...

docs/Off-line-Signatures-QR-Code.md
```diff
@@ -192,6 +211,55 @@ Generate the QR code from `offlineData`. Code example in Java:
...
+
+### 2. Computing Signatures on Mobile Device
+
+#### 2.1. Verify ECDSA Signature of Offline Data
+
+After user scans the QR code using a mobile app, the `ECDSA_SIGNATURE` should be validated on mobile device to verify authenticity of received data by taking contents of `offlineData` before the `ECDSA_SIGNATURE` and computing the ECDSA signature using `KEY_SERVER_PRIVATE`. Both signatures must match before continuing with the offline data signature verification.
```
romanstrobl: Use `ECDSA_QRDATA_SIGNATURE_BASE64 ` instead of `ECDSA_SIGNATURE` to avoid confusion?

docs/Off-line-Signatures-QR-Code.md
```diff
@@ -192,6 +211,55 @@ Generate the QR code from `offlineData`. Code example in Java:
...
+
+```java
+String signatureBaseString
+    = PowerAuthHttpBody.getSignatureBaseString(
+        "POST",
+        "/operation/authorize/offline",
+        BaseEncoding.base64().decode(nonce),
+        data.getBytes()
+);
+```
```
romanstrobl: We should add a link to the implementing class, it is hard to find:

https://github.com/wultra/powerauth-crypto/blob/master/powerauth-java-http/src/main/java/io/getlime/security/powerauth/http/PowerAuthHttpBody.java

docs/Off-line-Signatures-QR-Code.md
```diff
@@ -192,6 +211,55 @@ Generate the QR code from `offlineData`. Code example in Java:
...
+
+To verify signature, you need to call the SOAP method [`verifyOfflineSignature`](https://developers.wultra.com/docs/develop/powerauth-server/SOAP-Service-Methods#method-verifyofflinesignature) providing:
+
+- `signatureBaseString` (as obtained in 3.2.) - as data for verification
+- `signature` - value of the signature entered by the user (as obtained in 3.1., 2x8 digits)
+- `activationId` - identifier of the activation (to know which device is responsible for verification)
+- `signatureType` - type of the signature (`POSSESSION_KNOWLEDGE`).
 
-The `ECDSA_SIGNATURE` should be validated on mobile device to verify authenticity of received data by taking contents of `offlineData` before the `ECDSA_SIGNATURE` and computing the ECDSA signature using `KEY_SERVER_PRIVATE`. Both signatures must match before continuing with the offline data signature verification.
+The method returns information about signature verification, see the SOAP method documentation for details.
```
romanstrobl: Perhaps add comment about how to check the result of verification?

docs/Off-line-Signatures-QR-Code.md
petrdvorak: Good point.

docs/Off-line-Signatures-QR-Code.md
petrdvorak: Good point.

docs/Off-line-Signatures-QR-Code.md
petrdvorak: OK

docs/Off-line-Signatures-QR-Code.md
petrdvorak: Good point

docs/Off-line-Signatures-QR-Code.md
petrdvorak: Done.

PR #506, opened 2019-05-14
Update Off-line-Signatures-QR-Code.md

### LINKED ISSUE #503: Create the documentation on validating offline signature
Currently, we have a very vague description of how to validate offline signatures:

https://developers.wultra.com/docs/develop/powerauth-server/Offline-Signatures#verifying-offline-signatures

We should prepare a more detailed description, just like we have for generating offline codes:

https://developers.wultra.com/docs/develop/powerauth-webflow/Off-line-Signatures-QR-Code#how-to-generate-qr-codes

petrdvorak:
Fix #503: Create the documentation on validating the offline signature


petrdvorak:
@romanstrobl OK, I used an example with payment data.

PR #514, opened 2019-06-10
Fix #461: Enable consent screen and allow customization

### LINKED ISSUE #461: Enable consent screen and allow customization
Currently, we skip the consent screen. We should allow adding the consent screen back (either using a default OAuth 2.0 consent screen or - more likely - by implementing our own).

The consent screen should allow content customization (display the third party provider info, such as name, app name, company ID, or display the operation attributes) via Data Adapter.

romanstrobl:
I am opening a draft pull request to review the implementation.

Currently only documentation is missing, otherwise the pull request should be feature complete.


powerauth-webflow-authentication-consent/src/main/java/io/getlime/security/powerauth/lib/webflow/authentication/consent/controller/ConsentController.java
```diff
@@ -0,0 +1,258 @@
...
+     * Controller constructor.
+     * @param dataAdapterClient Data adapter client.
+     */
+    @Autowired
+    public ConsentController(DataAdapterClient dataAdapterClient) {
+        this.dataAdapterClient = dataAdapterClient;
+    }
+
+    /**
+     * Verifies the authorization code entered by user against code generated during initialization.
```
petrdvorak: Probably a copy-paste documentation.

powerauth-nextstep-model/src/main/java/io/getlime/security/powerauth/lib/nextstep/model/entity/ApplicationExtras.java
```diff
@@ -0,0 +1,71 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Extras for application context for OAuth 2.0 consent screen.
+ *
+ * @author Roman Strobl, roman.strobl@wultra.com
+ */
+public class ApplicationExtras {
```
petrdvorak: "Application extras" are supposed to be a free map with any information that 3rd party connectors may wish to add. In case we do typed attributes in `ApplicationExtras` class... there is really no sense to wrap them in `ApplicationExtras` class, these could be in `ApplicationContext` object directly. 

I see two ways to proceed:

- 1) Make `extras` attribute of `ApplicationContext` class a `Map<String, Object>`.
- 2) Make `extras` attribute `Map<String, Object>` as in 1) and if we feel that `requestedScopes` and `applicationOwner` are important application context attributes, let's put them to `ApplicationContext`

I am personally for option 1)...

powerauth-webflow/package.json
```diff
@@ -36,7 +36,8 @@
     "react-intl-redux": "^0.5.0",
     "hard-source-webpack-plugin": "^0.5.0-rc.2",
     "bluebird": "^3.5.0",
-    "intl": "^1.2.5"
+    "intl": "^1.2.5",
+    "sanitize-html": "^1.20.1"
```
petrdvorak: We should probably do some HTML sanitization server side as well...

powerauth-nextstep-model/src/main/java/io/getlime/security/powerauth/lib/nextstep/model/entity/ApplicationExtras.java
romanstrobl: I think 1) is fine.

powerauth-webflow/package.json
romanstrobl: Ok, I found this sanitizer: https://github.com/OWASP/java-html-sanitizer, we can try it.

romanstrobl:
I switched to `Map<String, Object>` for extras and wrote some more documentation.

I will look into server-side HTML sanitization.

PR #522, opened 2019-06-13
Fixes from testing of consent screen

romanstrobl:
I found several issues when testing consent screen with `POWERAUTH_TOKEN` step instead of `SMS_KEY`:
- Operation must not be returned in pending operations in case it was already processed (e.g. consent screen is displayed after `POWERAUTH_TOKEN` step)
- Removed invalid check for authentication method availability (the logic was broken but discovered only now that an additional step is available after `POWERAUTH_TOKEN`)
- Updated JavaDoc
- Added documentation of database changes
- Implemented HTML sanitization based on used policy in `sanitize-html` JavaScript project

### DIFF
--- a/docs/Web-Flow-0.22.0.md
+++ b/docs/Web-Flow-0.22.0.md
@@ -4,10 +4,20 @@
 
 ### Database Changes
 
-We added `NOT NULL` definitions to database tables to improve data integrity.
-
+Following database changes were introduced in version `0.22.0`:
+ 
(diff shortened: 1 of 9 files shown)


romanstrobl:
Note that we will have to change the check of `POWERAUTH_TOKEN` method due to 2FA login which also supports mobile token, but I would like to handle this functionality separately toghether with 2FA login implementation.

PR #532, opened 2019-06-24
Fix #526: Password encryption during transfer

### LINKED ISSUE #526: Password encryption during transfer
Add option to Web Flow to encrypt user password when received on backend using symetric cipher so that the password is encrypted during further transfer for verification against remote backends. The encryption algorithm, mode and padding should be configurable.

romanstrobl:
Note that this pull request is built on previous pull request because of required changes in Data Adapter model and interface: https://github.com/wultra/powerauth-webflow/pull/531

This pull request adds optional user password encryption during transfer using AES.


powerauth-webflow-authentication-form/src/main/java/io/getlime/security/powerauth/lib/webflow/authentication/method/form/controller/FormLoginController.java
```diff
@@ -99,7 +109,21 @@ protected AuthenticationResult authenticate(UsernamePasswordAuthenticationReques
...
-            final ObjectResponse<AuthenticationResponse> authenticateResponse = dataAdapterClient.authenticateUser(request.getUsername(), request.getPassword(), request.getOrganizationId(), operationContext);
+            final ObjectResponse<UserDetailResponse> lookupResponse = dataAdapterClient.lookupUser(request.getUsername(), request.getOrganizationId(), operationContext);
+
+            AuthenticationType authenticationType = configuration.getAuthenticationType();
+            String cipherTransformation = configuration.getCipherTransformation();
+            String passwordToSend;
+            if (authenticationType == AuthenticationType.BASIC) {
+                // Password is sent in plain text
+                passwordToSend = request.getPassword();
+            } else {
```
petrdvorak: I think that we should have `else-if` here (maybe rising exception if some other value is received?).

docs/Web-Flow-Configuration.md
```diff
@@ -67,7 +67,35 @@ powerauth.webflow.service.applicationEnvironment=
...
 powerauth.webflow.offlineMode.available=true
+
+# Configuration of Android Security Warning
+powerauth.webflow.android.showSecurityWarning=true
+```
+
+Encryption of user passwords during transport can be configured using following properties:
 ```
+# Configuration of Password Encryption
+powerauth.webflow.authentication.type=PASSWORD_ENCRYPTION_AES
```
petrdvorak: I am not sure about this property naming - I see why is this later in the code, but maybe having:

```
powerauth.webflow.password.encryption.type=...
```

... would make better sense from configuration perspective?

powerauth-webflow-authentication-form/src/main/java/io/getlime/security/powerauth/lib/webflow/authentication/method/form/service/PasswordEncryptionService.java
```diff
@@ -0,0 +1,116 @@
...
+ * @author Roman Strobl, roman.strobl@wultra.com
+ */
+@Service
+public class PasswordEncryptionService {
+
+    private static final Logger logger = LoggerFactory.getLogger(PasswordEncryptionService.class);
+
+    private final WebFlowServicesConfiguration configuration;
+
+    private final AESEncryptionUtils aesEncryptionUtils = new AESEncryptionUtils();
```
petrdvorak: ... I like calling this instance just `aes` - it makes the code more readable later.

powerauth-webflow-authentication-form/src/main/java/io/getlime/security/powerauth/lib/webflow/authentication/method/form/service/PasswordEncryptionService.java
```diff
@@ -0,0 +1,116 @@
...
+import javax.crypto.SecretKey;
+import java.nio.charset.StandardCharsets;
+
+/**
+ * Service for encryption of user passwords.
+ *
+ * @author Roman Strobl, roman.strobl@wultra.com
+ */
+@Service
+public class PasswordEncryptionService {
```
petrdvorak: I am actually not sure if we should have this as a "service", maybe a strategy pattern would work better for this. The reason is that this current way, we basically initialize everything for encryption even if it is not needed.

```java
interface PasswordProtection {
    String protect(String password);
}

class NoPasswordProtection implements PasswordProtection {
}

class AesEncryptionPasswordProtection implements PasswordProtection {
}
```

powerauth-webflow-authentication/src/main/java/io/getlime/security/powerauth/lib/webflow/authentication/controller/AuthMethodController.java
```diff
@@ -259,7 +259,7 @@ protected UpdateOperationResponse authorize(String operationId, String userId, S
     protected UpdateOperationResponse failAuthorization(String operationId, String userId, List<KeyValueParameter> params) throws NextStepServiceException, AuthStepException {
         logger.info("Fail step started, operation ID: {}, user ID: {}, authentication method: {}", operationId, userId, getAuthMethodName().toString());
         // validate operation before requesting update
-        GetOperationDetailResponse operation = validateOperationState(operationId);
+        GetOperationDetailResponse operation = getOperation(operationId);
```
petrdvorak: Is this really OK change?

powerauth-webflow-authentication/src/main/java/io/getlime/security/powerauth/lib/webflow/authentication/controller/AuthMethodController.java
```diff
@@ -291,7 +291,7 @@ protected UpdateOperationResponse failAuthorization(String operationId, String u
     protected UpdateOperationResponse cancelAuthorization(String operationId, String userId, OperationCancelReason cancelReason, List<KeyValueParameter> params) throws NextStepServiceException, AuthStepException {
         logger.info("Step cancel started, operation ID: {}, authentication method: {}", operationId, getAuthMethodName().toString());
         // validate operation before requesting update
-        GetOperationDetailResponse operation = validateOperationState(operationId);
+        GetOperationDetailResponse operation = getOperation(operationId);
```
petrdvorak: dtto

powerauth-webflow-authentication/src/main/java/io/getlime/security/powerauth/lib/webflow/authentication/controller/AuthMethodController.java
romanstrobl: Yes, the original method was duplicate and actually didn't have all functionality of getOperation().

powerauth-webflow-authentication/src/main/java/io/getlime/security/powerauth/lib/webflow/authentication/controller/AuthMethodController.java
romanstrobl: dtto

docs/Web-Flow-Configuration.md
romanstrobl: We can change it, no problem.

powerauth-webflow-authentication-form/src/main/java/io/getlime/security/powerauth/lib/webflow/authentication/method/form/service/PasswordEncryptionService.java
romanstrobl: Yes, this is nicer.

PR #594, opened 2019-07-29
Fix #593: Improve Android warning UI

### LINKED ISSUE #593: Improve Android warning UI
The current Android warning looks quite ugly, let's improve the CSS.

romanstrobl:
An attemt to make the warning less ugly - original UI:

image

After removing the Jumbotron:
image


romanstrobl:
In the new Wultra design:

image

PR #636, opened 2019-09-12
Fix #628: Implement consent engine

### LINKED ISSUE #628: Implement TPP registry and consent engine
To provide better boilerplate for PSD2 API solution, we should have own implementation of:

- *TPP Registry* - Database with information about 3rd parties.
- *Consent Engine* - Database with information about consent given to the third parties.

petrdvorak:
In this change, we deliver a base version of the consent engine. Consent engine is responsible for managing consents users gave to applications of the third parties. There are two new API contexts:

- `/consent` - Information about consents (consent templates) stored in the systems.
- `/user/consent` - Information about consents given by the system users.

### DIFF
--- a/powerauth-tpp-engine/src/main/java/io/getlime/security/powerauth/app/tppengine/repository/ConsentRepository.java
+++ b/powerauth-tpp-engine/src/main/java/io/getlime/security/powerauth/app/tppengine/repository/ConsentRepository.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2019 Wultra s.r.o.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getlime.security.powerauth.app.tppengine.repository;
+
+import io.getlime.security.powerauth.app.tppengine.repository.model.entity.ConsentEntity;
+import org.springframework.data.jpa.repository.Query;
+import org.springframework.data.repository.CrudRepository;
+import org.springframework.data.repository.query.Param;
+import org.springframework.stereotype.Repository;
+
+import java.util.Optional;
+
+/**
+ * Repository used for accessing the consent database.
+ *
+ * @author Petr Dvorak, petr@wultra.com
+ */
+@Repository
+public interface ConsentRepository extends CrudRepository<ConsentEntity, String> {
+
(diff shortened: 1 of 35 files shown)


powerauth-tpp-engine/src/main/java/io/getlime/security/powerauth/app/tppengine/repository/ConsentRepository.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Repository used for accessing the consent database.
+ *
+ * @author Petr Dvorak, petr@wultra.com
+ */
+@Repository
+public interface ConsentRepository extends CrudRepository<ConsentEntity, String> {
+
+    @Query("SELECT c FROM ConsentEntity c WHERE c.id = :id ORDER BY c.version DESC")
```
romanstrobl: Again, my preference would be to use `consent_id` instead of `id`. But this is really a matter of style / philosophy.

powerauth-tpp-engine/src/main/java/io/getlime/security/powerauth/app/tppengine/service/UserConsentService.java
```diff
@@ -0,0 +1,400 @@
...
+    /**
+     * Convert Map to JSON String.
+     * @param parameters Map.
+     * @return JSON String.
+     */
+    private String convertToJsonString(Map<String, String> parameters) {
+        try {
+            return new ObjectMapper().writeValueAsString(parameters);
+        } catch (JsonProcessingException e) {
+            return null;
```
romanstrobl: We should probably log an error here.

powerauth-tpp-engine/src/main/resources/application.properties
```diff
@@ -0,0 +1,38 @@
...
+
+# Database Configuration - MySQL
+spring.datasource.url=jdbc:mysql://localhost:3306/powerauth
+spring.datasource.username=powerauth
+spring.datasource.password=
+spring.datasource.driver-class-name=com.mysql.jdbc.Driver
+spring.jpa.properties.hibernate.dialect=org.hibernate.dialect.MySQL8Dialect
+spring.jpa.properties.hibernate.connection.characterEncoding=utf8
+spring.jpa.properties.hibernate.connection.useUnicode=true
+spring.jpa.hibernate.use-new-id-generator-mappings=false
```
romanstrobl: I would move this property to the bottom of this file, otherwise someone may comment it out together with MySQL properties.

powerauth-tpp-engine/src/test/java/io/getlime/security/powerauth/app/tppengine/TppEngineApplicationTests.java
```diff
@@ -0,0 +1,27 @@
...
+
+package io.getlime.security.powerauth.app.tppengine;
+
+import org.junit.runner.RunWith;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.junit4.SpringRunner;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest
+public class TppEngineApplicationTests {
```
romanstrobl: Do we need empty test runner?

PR #654, opened 2019-09-16
Fix #653: At least one redirect_uri must be registered with the client

### LINKED ISSUE #653: At least one redirect_uri must be registered with the client
After recent upgrade of Spring OAuth 2 integration we need to configure the `web_server_redirect_uri` parameter for the client.

See error:
```
Handling OAuth2 error: error="invalid_request", error_description="At least one redirect_uri must be registered with the client."
```

romanstrobl:


romanstrobl:
Ok, I will keep the pull request open until it is documented.

PR #662, opened 2019-09-16
Fix #661: Add requested OAuth 2.0 scopes into a strongly typed attribute

### LINKED ISSUE #661: Add requested OAuth 2.0 scopes into a strongly typed attribute
Currently, we only pass requested OAuth 2.0 scopes via weakly typed `extras` (as `_requestedScopes` key). However, this seems not very practical - the application might need to work with the scopes in a structured way.

petrdvorak:


petrdvorak:
I added a note into the migration guide.

PR #676, opened 2019-09-18
Fix #675: Filter IPv6 IP addresses when detecting client IP address

### LINKED ISSUE #675: Filter IPv6 IP addresses when detecting client IP address
Since AFS products such as ThreatMark require IPv4 IP addresses, we need to filter out IPv6 IP addresses when detecting client IP addresses.

romanstrobl:
We have to handle the case when client is IPv6 behind a proxy which forwards the IPv6 address correctly, however AFS cannot handle it because it only supports IPv4. I added an option to enable forcing of IPv4 and fallback logic in case no IPv4 address could not be determined.

### DIFF
--- a/powerauth-webflow-authentication/src/main/java/io/getlime/security/powerauth/lib/webflow/authentication/interceptor/WebSocketHandshakeInterceptor.java
+++ b/powerauth-webflow-authentication/src/main/java/io/getlime/security/powerauth/lib/webflow/authentication/interceptor/WebSocketHandshakeInterceptor.java
@@ -36,8 +36,8 @@ public class WebSocketHandshakeInterceptor implements HandshakeInterceptor {
                 if (ipAddress == null || !isIpv4Address(ipAddress)) {
                     ipAddress = servletRequest.getServletRequest().getRemoteAddr();
                     if (!isIpv4Address(ipAddress)) {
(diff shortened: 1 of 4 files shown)


powerauth-webflow-authentication/src/main/java/io/getlime/security/powerauth/lib/webflow/authentication/interceptor/WebSocketHandshakeInterceptor.java
```diff
@@ -25,9 +31,20 @@ public boolean beforeHandshake(@NonNull ServerHttpRequest request, @NonNull Serv
...
             String ipAddress = servletRequest.getServletRequest().getHeader("X-FORWARDED-FOR");
-            if (ipAddress == null) {
+            if (forceIpv4) {
+                // IPv4 logic
+                if (ipAddress == null || !isIpv4Address(ipAddress)) {
+                    ipAddress = servletRequest.getServletRequest().getRemoteAddr();
+                    if (!isIpv4Address(ipAddress)) {
+                        // IP address is null in case IPv4 address could not be determined, it should not be sent to AFS
+                        ipAddress = null;
+                    }
```
saalistaja: Please add logging of the `else` case

PR #678, opened 2019-09-24
Fix #677: AFS: configuration of IP address detection

### LINKED ISSUE #677: AFS: configuration of IP address detection
In case IP address detection is unreliable, it should be possible to turn it off. I suggest to make it the default, as this is a typical scenario in PSD2 internal deployments until the infrastructure forwards the IP address via HTTP headers.

romanstrobl:
As per request from the client, we will make it possible to disable client IP address detection.

### DIFF
--- a/powerauth-webflow-authentication/src/main/java/io/getlime/security/powerauth/lib/webflow/authentication/configuration/WebFlowServicesConfiguration.java
+++ b/powerauth-webflow-authentication/src/main/java/io/getlime/security/powerauth/lib/webflow/authentication/configuration/WebFlowServicesConfiguration.java
@@ -106,13 +106,13 @@ public class WebFlowServicesConfiguration {
      * Whether client IP address is detected in anti-fraud system integration.
      */
     @Value("${powerauth.webflow.afs.detectIpAddress:false}")
-    private boolean afsDetectIpAddress;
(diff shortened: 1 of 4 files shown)


powerauth-webflow-authentication/src/main/java/io/getlime/security/powerauth/lib/webflow/authentication/configuration/WebFlowServicesConfiguration.java
```diff
@@ -223,6 +229,14 @@ public AfsType getAfsType() {
         return afsType;
     }
 
+    /**
+     * Get whether client IP address is detected in anti-fraud system integration.
+     * @return Whether client IP address is detected in anti-fraud system integration.
+     */
+    public boolean getAfsDetectIpAddress() {
```
saalistaja: I would recommend to use `isXXX` or `hasXXX` for a boolean getter. Maybe change the field name a little bit?

`afsIpAddressDetectionEnabled` (too long)

romanstrobl: This variable name is too long.

PR #690, opened 2019-10-11
Fix #563: Improve handling for blocked accounts in SCA login and user profile

### LINKED ISSUE #563: Improve handling for blocked accounts in SCA login and user profile
Currently a `null` value is returned for `userId` whose account is blocked in first step of SCA login (method `lookupUser()` from Data Adapter).

It would be better to return an enum such as `UserAccountStatus` with two options: `ACTIVE` and `NOT_ACTIVE`. 

In case the user account is `ACTIVE`, the authentication is done using regular authentication flow.

In case the user account is `NOT_ACTIVE`, the `userId`, `organizationId` would be normally persisted for the operation, however a fake sms and password form would be shown. This is currently already done for `null` value of `userId`, however the `null` value is used as a side effect, so this behaviour is not transparent.

Furthermore, the `createAuthorizationSms()` method in Data Adapter would contain the new `UserAccountStatus` enum so that Data Adapter can decide whether to send the SMS or not based on value of this parameter, the text of the SMS message can also be updated if necessary. The `createAuthorizationSms()` method would also receive a proper `userId` and `organizationId` instead of `null` values which are currently sent which allows better handling of SMS delivery.

Finally the `fetchUserDetail()` method would implement the same logic, so that when user account gets blocked during the operation, the OAuth resource server would not provide user information for a `NOT_ACTIVE` account.

romanstrobl:


romanstrobl:

PR #691, opened 2019-10-14
Fix #533: Timed out operation is not updated in Next Step

### LINKED ISSUE #533: Timed out operation is not updated in Next Step
When an operation times out in `MobileTokenOnlineController`, Next Step is not informed about the change of the operation. The Next Step should be called with information about updated operation state by calling `nextStepClient.updateOperation()` with following information:
```
            request.setAuthStepResult(AuthStepResult.AUTH_METHOD_FAILED);
            response.setTimestampExpires(operation.getTimestampExpires());
            response.setResult(AuthResult.FAILED);
            response.setResultDescription("operation.timeout");
```

The same notification should be handled in `checkOperationExpiration()` method and Data Adapter should be notified about failed operation similarly as in `failAuthorization()`.

romanstrobl:
`MobileTokenOnlineController` has custom timeout handling due to sending of push messages, so the `cancelAuthorization` method needs to be called so that:
- Next Step operation status is updated
- Data Adapter is notified about operation change
- AFS receives information about operation timeout


romanstrobl:

PR #692, opened 2019-10-14
Fix #574: Store information about used authorization instrument

### LINKED ISSUE #574: Store information about used authorization instrument
Currently it is non-trivial to find out which authorization instrument was used in given step (mobile token online/offline, SMS key, HW key, etc.).

We should store information about used authorization instrument in a new column for each operation history item.

romanstrobl:
This pull request introduces logging of used authentication / authorization instrument(s) into operation history, so that it is easier to find out which authentication / authorization instruments were used during individual steps.


romanstrobl:

PR #693, opened 2019-10-14
Fix #573: Disable default operation

### LINKED ISSUE #573: Enable customization of a default operation
Using the default login operation is problematic in multiple way (different URL, page refresh issues, hardcoded operation name, hardcoded formData). We should disable default operation and return an error instead.

Fix resolves: https://github.com/wultra/powerauth-webflow/issues/566

romanstrobl:
@petrdvorak please check that this functionality is not needed in any new projects that are currently running.


romanstrobl:
We'll need to call Data Adapter to create the operation based on `client_id` and `scope` parameters. Closing this pull request.

PR #700, opened 2019-10-29
Fix #699: Two-step selection of account breaks mobile token

### LINKED ISSUE #699: Two-step selection of account breaks mobile token
I received a push notification when I was about to select an account from the combo box. I confirmed the operation on mobile, nothing happened on the web and then, after selecting an account, I was prompted for SMS.

Steps to reproduce:

1. Have a user with a mobile token set up.
2. Create a payment.
3. Go through the login (on web and mobile).
=> You are now on the screen with an account selection.
=> You received push message on mobile.

4. Confirm operation on mobile.
5. Select the account.
=> You are prompted for SMS auth.

romanstrobl:
I fixed the issue, however I cannot test it properly because I don't have a USB-C to USB adaper / USB-C to lightning cable during my trip. So I just tested that the `mobile_token_active` value is written into the DB which should cause the pending operations endpoint start returning correct response at the right moment. I will test it more thoroughly tomorrow once I buy the cable.

### DIFF
--- a/docs/Web-Flow-0.23.0.md
+++ b/docs/Web-Flow-0.23.0.md
@@ -30,6 +30,7 @@ CREATE INDEX wf_operation_hash ON wf_operation_session (operation_hash);
 CREATE INDEX wf_websocket_session ON wf_operation_session (websocket_session_id);
 
 ALTER TABLE ns_operation_history ADD request_auth_instruments VARCHAR(256);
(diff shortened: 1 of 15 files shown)


romanstrobl:
I have verified that the reported scenario can no longer be reproduced after the fix in mobile otken. Without the fix I can get into the state reported in the issue.

romanstrobl:
I am merging the pull request so that we have enough time to test this functionality before release.

PR #702, opened 2019-11-01
Update Web Flow documentation

romanstrobl:
Update of Web Flow documentation, mainly regarding the AFS implementation.


romanstrobl:
I will merge this pull request shortly, so that up-to-date documentation is available for our clients for integration purposes. In case there are any issues let's fix them separately.

PR #736, opened 2019-12-16
Fix #596: Consent screen should allow longer consent texts

### LINKED ISSUE #596: Consent screen should allow longer consent texts
In case the consent text is long and contains multiple options, the user needs to use a scrollbar. The UI should be improved to support longer consent forms. The consent form still needs to be displayed without horizontal scrolling both on a computer and on a mobile phone which has much smaller width.

romanstrobl:
When the consent HTML has more than configured number of characters a wide consent panel is displayed to allow better display of large consent texts.

### DIFF
--- a/powerauth-webflow-client/src/main/resources/static/css/main.css
+++ b/powerauth-webflow-client/src/main/resources/static/css/main.css
@@ -197,6 +197,6 @@ a {
     float: left;
 }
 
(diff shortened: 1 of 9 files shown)


romanstrobl:
Documentation is pending, I will prepare it tomorrow.

PR #771, opened 2020-01-27
Fix #769: Improve validation of operation against HTTP session ID

### LINKED ISSUE #769: Improve validation of operation against HTTP session ID
We should improve the way we load operations in another session than the first one that the operation was opened in. Currently, we rely on a WebSocket session - we should also incorporate the additional HTTP session check.

romanstrobl:
HTTP session ID is checked when operation is initialized or continued.

I also tightened the bolts during Web Socket registration, Web Socket session can be registered only once.

### DIFF
--- a/powerauth-webflow-authentication/src/main/java/io/getlime/security/powerauth/lib/webflow/authentication/service/OperationSessionService.java
+++ b/powerauth-webflow-authentication/src/main/java/io/getlime/security/powerauth/lib/webflow/authentication/service/OperationSessionService.java
@@ -60,12 +60,19 @@ public class OperationSessionService {
      * @param operationId Operation ID.
      * @param httpSessionId HTTP session ID.
      * @param result Operation result.
+     * @return Whether HTTP session was successfully registered for the operation.
      */
-    public void persistOperationToSessionMapping(String operationId, String httpSessionId, AuthResult result) {
+    public boolean registerHttpSession(String operationId, String httpSessionId, AuthResult result) {
+        OperationSessionEntity existingSession = getOperationToSessionMapping(operationId);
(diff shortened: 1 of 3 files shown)


powerauth-webflow-authentication/src/main/java/io/getlime/security/powerauth/lib/webflow/authentication/service/OperationSessionService.java
```diff
@@ -152,8 +159,9 @@ public boolean registerWebSocketSession(String operationHash, String webSocketSe
             return false;
         }
         String existingWebSocketSessionId = operationSessionEntity.getWebSocketSessionId();
-        if (existingWebSocketSessionId != null && !webSocketSessionId.equals(existingWebSocketSessionId)) {
-            // Registration failed because operation is being accessed from another browser tab or window
+        if (existingWebSocketSessionId != null) {
```
petrdvorak: Just a question - why did we change this condition? Does this not this cause operation failure during step change? (websocket ID exists and is the one it is supposed to be)

romanstrobl: The change was done to allow Web Socket registration only once in `startHandshake.js`. We do not need to allow it multiple times. When step changes the Web Socket session ID does not change, the Web Socket is opened only once during initial handshake. The only case when Web Socket would need to re-register would be page refresh which is no longer supported:
  - in case operation ID is known, the operation fails after refresh when user overrides the refresh confirmation dialog
  - in case of implicit operation a new operation is started which can register new Web Socket session because the operation ID is different

PR #836, opened 2020-08-31
Fix #835: Set Maven Compiler Source And Target to 8

### LINKED ISSUE #835: Set maven.compiler.target to 8
WebFlow compilation should produce binaries which still run on Java 8, although they are compiled with Java 11.

romanstrobl:
Let's revert back to Java 8 for the source and target.


petrdvorak (approved):
Looks OK

PR #840, opened 2020-09-02
Fix #839: Inconsistent nullability in DDL for external_id

### LINKED ISSUE #839: Inconsistent nullability in DDL for external_id
The `external_id` should not be defined as `NOT NULL`, which is not the case for all supported databases.

romanstrobl:


petrdvorak (approved):
Looks OK

PR #877, opened 2020-09-18
Fix #843: Add status endpoint for TPP engine

### LINKED ISSUE #843: Add status endpoint for TPP engine
Service status endpoint is missing for TPP engine. We should add it for consistency with other applications.

romanstrobl:


petrdvorak (approved):
Looks OK, thank you for spotting this is missing...

PR #894, opened 2020-10-07
Fix version in pom.xml file and allow build with Java 8

petrdvorak:


romanstrobl (approved):
Looks good, I tried to compile code both with Java 8 and Java 11 and it works fine.

PR #904, opened 2020-11-16
Fix #903: Do not log tokenId in revokeToken

### LINKED ISSUE #903: Do not log tokenId in revokeToken
A static analysis tool report contains information that the `tokenId` should not be logged in `revokeToken` method. Let's remove it.

romanstrobl:


petrdvorak (approved):
Looks OK

PR #920, opened 2020-12-08
Handle Push client initialization error

romanstrobl:


powerauth-webflow/src/main/java/io/getlime/security/powerauth/app/webflow/configuration/PowerAuthWebServiceConfiguration.java
```diff
@@ -90,12 +91,17 @@ public PowerAuthClient powerAuthClient() {
...
     @Bean
     public PushServerClient pushServerClient() {
-        PushServerClient client = new PushServerClient(powerAuthPushServiceUrl);
-        // whether invalid SSL certificates should be accepted
-        if (acceptInvalidSslCertificate) {
-            sslConfigurationService.trustAllCertificates();
+        try {
+            PushServerClient client = new PushServerClient(powerAuthPushServiceUrl);
+            // whether invalid SSL certificates should be accepted
+            if (acceptInvalidSslCertificate) {
```
petrdvorak, line 97: This is very likely a historical leftover code from the Unirest times - the invalid SSL certificates should be handled on the `PushServerClient` level...

petrdvorak (approved):
Looks OK but maybe we should have a look if we handle the invalid SSL certificates correctly.

powerauth-webflow/src/main/java/io/getlime/security/powerauth/app/webflow/configuration/PowerAuthWebServiceConfiguration.java
romanstrobl: Good catch, removed.

PR #933, opened 2021-01-11
Fix #610: Fix color of spinner in webflow

### LINKED ISSUE #610: Fix color of spinner in webflow
It looks bad on dark background

romanstrobl:
The Spinner color can be customized, it is now white in Wultra customization:
image


kober32 (approved):
Looks OK

PR #934, opened 2021-01-11
Fix #917: Fix CSS issue after migration to React 16

### LINKED ISSUE #917: Fix CSS issue after migration to React 16
The styling is broken after migration to React 16.

romanstrobl:
The `panel-body` class is no longer used by `Panel` in React 16, so I switched the customization to the `panel-default` class. I visually compared the `0.24.1` version and the fix and they look the same.


kober32 (approved):
👍🏾

PR #952, opened 2021-01-28
Model updates for identity and password / OTP authentication

romanstrobl:
The pull request contains following changes:
- new request/response classes for Next Step identity and authentication endpoints
- updates of existing request/response classes for easier integration
- migration of model classes to Lombok
- stubs of new REST controllers
- version updated to `1.1.0-SNAPSHOT`


petrdvorak (approved):
Looks OK, if we meet some missing attributes, let's add it later.

PR #956, opened 2021-03-11
First version of Next Step identity

romanstrobl:
This pull request contains first version of Next Step identity implementation. 

Commits include:
- identity configuration (applications, OTP/credential policies and definitions, hashing configuration, role definitions, etc.)
- user identity management (users, credentials, contacts, roles, aliases)
- credential and OTP services including management of counters, password generation and validation using Passay library
- authentication services (credential, OTP, credential + OTP)
- integration of authentication with Next Step operations
- Data Adapter and Next Step clients migrated to Rest Client
- partial migration of Data Adapter calls to Next Step used as a proxy
- tests for authentication, generating and validating credentials, generating OTP, and managing user identity lifecycle
- record encryption, hashing and end-to-end encryption

Out of scope (will be delivered later):
- authentication method downgrade and method choice for PA token operations - implemented in https://github.com/wultra/powerauth-webflow/pull/958
- improved integration with PA server ("small" operations) - implemented in https://github.com/wultra/powerauth-webflow/pull/960
- migrating the rest of Data Adapter calls to Next Step proxy where it makes sense
- extension of user info endpoints
- logging, auditing, code cleanup, more tests, PostgreSQL DDL, documentation, migration guide - partially implemented in https://github.com/wultra/powerauth-webflow/pull/961

PR #958, opened 2021-03-31
Next Step identity - additional features

romanstrobl:
This pull request contains another round of features on top of #956. I am merging this pull request into branch `feature/next-step-identity` to allow reviewing these smaller changes separately. Then we can merge all changes into `develop`.

Changes include:
- organization is linked in DB from credential definition instead from application to allow configuration with a single applications containing multiple organizations and multiple credentials
- large varchars are limited to 4000 characters due to Oracle limit
- option to disable e2e encryption for generated temporary credentials
- customization of operation expiration per operation name
- allow larger extras in DB
- configuration of maximum number of failures per operation name / authentication method
- temporary credentials should not be usable after their expiration
- avoid unnecessary credential verification when OTP is not active
- authentication method downgrade in update operation
- authentication method choice in update operation
- check activation status in PowerAuth server when obtaining pending operations and enabled authentication methods
- hashing and encryption support in credential history
- remove temporary workaround for login operation
- various small bugfixes


bodynek (approved):
Looks good, comments in Slack

petrdvorak (approved):
Looks OK

PR #960, opened 2021-04-05
Integrate PowerAuth operations into Next Step and Web Flow

romanstrobl:
This pull request integrates PowerAuth operations into Next Step and Web Flow:
- PA operation support is optional, it needs to be enabled in Next Step and Web Flow
- choice of `POWERAUTH_TOKEN` method or mobile token status change in SCA methods triggers creating a new PowerAuth operation
- PA operation ID is stored in operation history
- pending operations are checked for expiration in PA server and expiration time is updated if needed, polling triggered by Web Flow polling is used, callbacks are not provided by PA server yet
- Web Flow mobile API endpoints handle approval, fail approval, and reject requests
- mobile token offline mode also handles PA operation updates
- operations canceled in Web Flow by the user trigger PA operation rejection
- canceled operations due to other reasons (e.g. browser window closed, unexpected errors) trigger a cancel request
- when PA operation calls fail, Next Step operation is failed
- when PA operations expire, Next Step operations are canceled

I suggest we keep PA operations disabled by default in Next Step and Web Flow for this release, because:
- frequent polling for expiration consumes resources of PA server
- mobile token application is not fully ready for custom number of attempts in PA operation
- we need to test this feature properly, it complicates Next Step and Web Flow states

### DIFF
--- a/docs/sql/mysql/create_schema.sql
+++ b/docs/sql/mysql/create_schema.sql
@@ -437,6 +437,7 @@ CREATE TABLE ns_operation_history (
   chosen_auth_method          VARCHAR(32),                         -- Information about which authentication method was chosen, in case user can chose the authentication method.
   mobile_token_active         BOOLEAN NOT NULL DEFAULT FALSE,      -- Information about if mobile token is active during the particular authentication step, in order to show the mobile token operation at the right time.
   authentication_id           VARCHAR(256),                        -- Reference to the authentication record.
+  pa_operation_id             VARCHAR(256),                        -- PowerAuth operation ID for PowerAuth operations.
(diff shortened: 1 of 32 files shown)


romanstrobl:
The problem with expiration of operations in mobile token was actually a bug in mobile token: https://github.com/wultra/mtoken-sdk-ios/issues/49 Once fixed, operations should properly expire with PA operation expiration time, in case  it is earlier than NS operation expiration time.

bodynek (approved):
didn't find a problem

romanstrobl:
@petrdvorak Can I merge this feature? I'd like to update `develop` to be in par with the snapshot release of Next Step identity for further development. Thanks.

PR #961, opened 2021-04-06
Updates for release 1.1.0-SNAPSHOT

romanstrobl:
This pull request contains changes included in release `1.1.0-SNAPSHOT`:
- Logging in controllers and sensitive parts of the application.
- Timeout checks disabled in mobile token component to allow push message delivery triggered by authentication polling.
- Simplified JPA entities and repositories - instead of using many repositories with `@ManyToOne` mappings (which turned out to be problematic in Oracle due to JPA entity synchronization issues) I switched to bidirectional JPA mappings for user identity and its children. The JPA code is simpler now but it will require some more testing because it is a big conceptual change.


powerauth-nextstep/src/main/java/io/getlime/security/powerauth/app/nextstep/controller/AuthMethodController.java
```diff
@@ -187,7 +188,10 @@ public AuthMethodController(AuthMethodService authMethodService) {
      */
     @RequestMapping(value = "user/auth-method/enabled/list", method = RequestMethod.POST)
     public ObjectResponse<GetEnabledMethodListResponse> getEnabledMethodList(@Valid @RequestBody ObjectRequest<GetEnabledMethodListRequest> request) throws InvalidConfigurationException {
+        // Log level is FINE to avoid flooding logs, this endpoint is used all the time.
```
bodynek, line 191: 👍

powerauth-nextstep/src/main/java/io/getlime/security/powerauth/app/nextstep/service/CredentialCounterService.java
```diff
@@ -95,7 +95,6 @@ public UpdateCounterResponse updateCredentialCounter(UpdateCounterRequest reques
      * @throws InvalidRequestException Thrown when request is invalid.
      */
     public void updateCredentialCounter(CredentialEntity credential, AuthenticationResult authenticationResult) throws InvalidRequestException {
-        credential.setAttemptCounter(credential.getAttemptCounter() + 1);
```
bodynek, line 98: where did this goes?

powerauth-nextstep/src/main/java/io/getlime/security/powerauth/app/nextstep/service/UserIdentityService.java
```diff
@@ -218,8 +208,9 @@ public CreateUserResponse createUserIdentity(CreateUserRequest request) throws U
             List<UserContactDetail> addedContacts = updateContacts(user, contacts);
             response.getContacts().addAll(addedContacts);
         }
-        // Save user identity snapshot to the history table
-        saveUserIdentityHistory(user);
+        // Save user identity and a snapshot to the history table
+        updateUserIdentityHistory(user);
```
bodynek, line 212: shouldn't it be reversed?
First userIdentity and then its History to be hooked on the userIdentity?

bodynek (changes requested):
just few small tips, otherwise looks good

powerauth-nextstep/src/main/java/io/getlime/security/powerauth/app/nextstep/service/CredentialCounterService.java
romanstrobl: This call was duplicate, check other usages of `credential.setAttemptCounter()`. The attempt counter is increased earlier during authentication right after request validation.

powerauth-nextstep/src/main/java/io/getlime/security/powerauth/app/nextstep/service/UserIdentityService.java
romanstrobl: The user identity history no longer uses a dedicated repository. So the history is updated (added to the user identity) and then it is saved by saving the user identity. The `updateUserIdentityHistory` method call does not call any `save()` method. So, the `save()` on userIdentity needs to be always called last, otherwise the changes wouldn't be saved into database.

bodynek: oh, you're right. Then it's good

bodynek (approved):
looks good

powerauth-nextstep/src/main/java/io/getlime/security/powerauth/app/nextstep/service/CredentialService.java
```diff
@@ -73,9 +75,10 @@
...
      */
     @Autowired
-    public CredentialService(UserIdentityLookupService userIdentityLookupService, CredentialDefinitionService credentialDefinitionService, CredentialRepository credentialRepository, CredentialHistoryService credentialHistoryService, IdGeneratorService idGeneratorService, NextStepServerConfiguration nextStepServerConfiguration, CredentialGenerationService credentialGenerationService, CredentialValidationService credentialValidationService, CredentialProtectionService credentialProtectionService, EndToEndEncryptionService endToEndEncryptionService) {
+    public CredentialService(UserIdentityLookupService userIdentityLookupService, CredentialDefinitionService credentialDefinitionService, CredentialRepository credentialRepository, CredentialHistoryService credentialHistoryService, IdGeneratorService idGeneratorService, NextStepServerConfiguration nextStepServerConfiguration, CredentialGenerationService credentialGenerationService, CredentialValidationService credentialValidationService, CredentialProtectionService credentialProtectionService, EndToEndEncryptionService endToEndEncryptionService, UserIdentityRepository userIdentityRepository) {
```
petrdvorak, line 81: This is a candidate for some service catalogue. We are autowiring the universe here. We can leave it for now but let's have a look at this later. (I will open an issue.)

powerauth-nextstep/src/main/java/io/getlime/security/powerauth/app/nextstep/service/CredentialService.java
```diff
@@ -624,9 +631,14 @@ public CredentialSecretDetail createCredential(UserIdentityEntity user, Credenti
...
-        credentialRepository.save(credential);
+        if (newCredential) {
+            // Credential needs to be added after all validations, otherwise JPA may save the credential prematurely
+            user.getCredentials().add(credential);
+        }
+
         // Save credential into credential history
-        credentialHistoryService.createCredentialHistory(credential, new Date());
+        credentialHistoryService.createCredentialHistory(user, credential, new Date());
+        userIdentityRepository.save(user);
```
petrdvorak, line 641: Should you assign to `user` here?

```java
user = userIdentityRepository.save(user);
```

You are working with the `user` variable later in the code...

petrdvorak (changes requested):
There is one place where you do not assign entity on save (despite using it later in the code). Also, our autowired contructors are getting out of hand by the number of parameters - I opened a separate issue on that. Finally, and I did not comment on it in this code review - it seems you are deliberately avoid the `final` keyword, we should use it for variables that are final in the context. I suggest running the code inspection in IntelliJ Idea at some point to identify simple to fix issues like missing `final` keyword.

powerauth-nextstep/src/main/java/io/getlime/security/powerauth/app/nextstep/service/CredentialService.java
romanstrobl: When I do this, IntelliJ IDEA marks the variable as unused, so probably not.

powerauth-nextstep/src/main/java/io/getlime/security/powerauth/app/nextstep/service/CredentialService.java
romanstrobl: Yes, I am aware of this issue.

romanstrobl:
Agree on the missing `final` keywords. Let's raise a separate issue.

romanstrobl:
Also good point about static analysis. I also need to run other static analysis tools such as Coverity scan, there was no time for that yet.

powerauth-nextstep/src/main/java/io/getlime/security/powerauth/app/nextstep/service/CredentialService.java
petrdvorak: You do use the variable on line 688, se below. This was not the case before (and so the variable was unused), but now it is the case...

petrdvorak: Ignore... I see this, there was one more collapse in the code review block... My bad.

petrdvorak (approved):
Looks OK

romanstrobl:
Thanks for the review. I will merge this PR after we review the previous pull requests.

PR #974, opened 2021-04-12
Fix #972: Migrate TPP client to Rest Client

### LINKED ISSUE #972: Migrate TPP client to Rest Client
The TPP engine should be migrated to Rest client from RestTemplate.

romanstrobl:


powerauth-tpp-engine-client/src/main/java/io/getlime/security/powerauth/lib/tpp/engine/client/TppEngineClient.java
```diff
@@ -31,89 +34,62 @@
...
+     * @param serviceUrl Base service URL.
+     * @throws TppEngineClientException Thrown when REST client initialization fails.
      */
-    public TppEngineClient(String serviceUrl, ObjectMapper objectMapper) {
-        this.serviceUrl = serviceUrl;
-        if (objectMapper != null) {
-            this.objectMapper = objectMapper;
-        } else {
-            this.objectMapper = objectMapper();
+    public TppEngineClient(String serviceUrl) throws TppEngineClientException {
```
petrdvorak, line 65: I think we need to keep an option to pass `ObjectMapper` here - maybe just introduce a wrapper method?

petrdvorak (approved):
Looks OK, I would just add constructor that allows setting a custom ObjectMapper...

powerauth-tpp-engine-client/src/main/java/io/getlime/security/powerauth/lib/tpp/engine/client/TppEngineClient.java
romanstrobl: We can use the constructor `TppEngineClient(RestClientConfiguration config)`, which allows setting the custom `ObjectMapper` as a parameter. I think this solves are customization use cases.

petrdvorak: 👍  Yes, this works for me.

PR #975, opened 2021-04-13
Fix #968: Scope used in demo application does not match configured scope for OAuth client

### LINKED ISSUE #968: Scope used in demo application does not match configured scope for OAuth client
The Web Flow demo application uses either `aisp` or `pisp` scope, however the table `oauth_client_details` contains only the `profile` scope. Thus, the OAuth 2.0 authentication produces an error during the redirect after successful authentication:

```
Location: http://localhost:8080/powerauth-webflow-client/connect/demo?error=invalid_scope&error_description=Invalid%20scope:%20aisp&state=a60f0f91-23f8-4ce9-a10b-500347b2240b&scope=profile
```

romanstrobl:


petrdvorak (approved):
Looks OK

PR #982, opened 2021-04-16
Fix #981: Update soft attempt counter reset logic

### LINKED ISSUE #981: Update soft failed attempt counter reset logic
Currently the soft failed attempt counter reset logic in endpoint `/credential/counter/reset-all` is following:
- when the credential status is `BLOCKED_TEMPORARY`, the soft failed attempt counter is reset and status is set to `ACTIVE`

We should update the logic:
- when the credential status is `BLOCKED_TEMPORARY`, the soft failed attempt counter is reset and status is set to `ACTIVE`
- when the credential status is `ACTIVE`, the soft failed attempt counter is reset

This will allow all clients to start with clean slate in the beginning of every day.

romanstrobl:
Updated logic for reseting soft counters based on more detailed requirements.

### DIFF
--- a/powerauth-nextstep/src/main/java/io/getlime/security/powerauth/app/nextstep/service/CredentialCounterService.java
+++ b/powerauth-nextstep/src/main/java/io/getlime/security/powerauth/app/nextstep/service/CredentialCounterService.java
@@ -133,8 +133,8 @@ public class CredentialCounterService {
      *
      * Method behavior depends on the counter reset mode:
      * <ul>
-     *     <li>RESET_ACTIVE_AND_BLOCKED_TEMPORARY - counters for credentials with ACTIVE and BLOCKED_TEMPORARY status are reset, status is set to ACTIVE</li>
-     *     <li>RESET_BLOCKED_TEMPORARY - counters for credentials with BLOCKED_TEMPORARY status are reset</li>
+     *     <li>RESET_BLOCKED_TEMPORARY - reset soft failed attempt counters for credentials with BLOCKED_TEMPORARY status, change status to ACTIVE</li>
+     *     <li>RESET_ACTIVE_AND_BLOCKED_TEMPORARY - reset soft failed attempt counters for credentials with ACTIVE and BLOCKED_TEMPORARY statuses, change status to ACTIVE if required</li>
      * </ul>
      * @param request Reset counters request.
      * @return Reset counters response.
@@ -144,13 +144,13 @@ public class CredentialCounterService {
     public ResetCountersResponse resetCounters(ResetCountersRequest request) throws InvalidRequestException {
         int resetCounter = 0;
         switch (request.getResetMode()) {
+            case RESET_BLOCKED_TEMPORARY:
+                resetCounter += credentialRepository.resetSoftFailedCountersForBlockedTemporaryStatus();
+                break;
+
             case RESET_ACTIVE_AND_BLOCKED_TEMPORARY:
                 resetCounter += credentialRepository.resetSoftFailedCountersForBlockedTemporaryStatus();
                 resetCounter += credentialRepository.resetSoftFailedCountersForActiveStatus();
-                break;
-
-            case RESET_BLOCKED_TEMPORARY:
-                resetCounter += credentialRepository.resetSoftFailedCountersForBlockedTemporaryStatus();
                 break;
 
             default:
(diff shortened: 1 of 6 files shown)


powerauth-nextstep/src/main/java/io/getlime/security/powerauth/app/nextstep/service/CredentialCounterService.java
```diff
@@ -137,14 +137,26 @@ public void updateCredentialCounter(CredentialEntity credential, AuthenticationR
     @Transactional
     public ResetCountersResponse resetCounters(ResetCountersRequest request) {
```
petrdvorak, line 138: The request should contain info about the type of reset. Otherwise, the side-effect can be unexpected. For example:

```json
{
    "active": true,
    "blocked": true
}
```

powerauth-nextstep/src/main/java/io/getlime/security/powerauth/app/nextstep/service/CredentialCounterService.java
```diff
@@ -137,14 +137,26 @@ public void updateCredentialCounter(CredentialEntity credential, AuthenticationR
...
+        // Reset soft counters for credentials with status BLOCKED_TEMPORARY and set status to ACTIVE
         for (CredentialEntity credential: blockedCredentials) {
             credential.setStatus(CredentialStatus.ACTIVE);
             credential.setFailedAttemptCounterSoft(0);
             credential.setTimestampBlocked(null);
+            resetCounter++;
         }
         credentialRepository.saveAll(blockedCredentials);
+        // Reset soft counters for credentials with status ACTIVE with non-zero soft counter
+        final List<CredentialEntity> activeCredentials = credentialRepository.findAllByStatus(CredentialStatus.ACTIVE);
```
petrdvorak, line 150: This is a horrible query that will basically return all customers at once. We should:

- scope it only to those credentials that have non-zero counter
- use paging / limit results (unless Hibernate does some serious magic)

powerauth-nextstep/src/main/java/io/getlime/security/powerauth/app/nextstep/service/CredentialCounterService.java
```diff
@@ -137,14 +137,26 @@ public void updateCredentialCounter(CredentialEntity credential, AuthenticationR
     @Transactional
     public ResetCountersResponse resetCounters(ResetCountersRequest request) {
         final List<CredentialEntity> blockedCredentials = credentialRepository.findAllByStatus(CredentialStatus.BLOCKED_TEMPORARY);
```
petrdvorak, line 139: Consider paging.

petrdvorak (approved):
The change is less terrible than I expected, but we should have a look at performance and granularity of configuration.

petrdvorak (changes requested):
Apologies - the approve was unintended.

powerauth-nextstep/src/main/java/io/getlime/security/powerauth/app/nextstep/service/CredentialCounterService.java
romanstrobl: I think we can avoid performance issues by executing SQL updates directly. In this case the logic is quite basic, so we don't need to query the data. Thanks for pointing it out.

powerauth-nextstep/src/main/java/io/getlime/security/powerauth/app/nextstep/service/CredentialCounterService.java
romanstrobl: I prefer an enumeration parameter because I think there are two use cases here, not four (from the booleans in example): `RESET_BLOCKED`, `RESET_ACTIVE_AND_BLOCKED`.

powerauth-nextstep/src/main/java/io/getlime/security/powerauth/app/nextstep/service/CredentialCounterService.java
romanstrobl: I will switch to SQL update to avoid performance issues.

romanstrobl:
All performance issues should be gone now.

petrdvorak (approved):
Looks OK

PR #984, opened 2021-04-19
Fix #964: Provide GET alternatives in Next Step REST API

### LINKED ISSUE #964: Provide GET alternatives in Next Step REST API
There are quite a few Next Step REST API methods which are used for listing and obtaining details which could provide a GET alternative to the POST method. We need to make sure the GET methods are used when no information is leaked in the URL.

romanstrobl:
I added `GET` method alternatives for REST API methods used for getting lists and details.


romanstrobl:
I attach the Swagger for an easier review.

We are getting a bit closer to how REST should be used:
- For `list` endpoints calling `GET` on a resource returns its list.
- For `detail` endpoints I sill left the `detail` word in the URL because using e.g. operation name in the URL could lead to a clash with other endpoints.
- Adding `DELETE` alternatives is out of scope, I am not sure if it is necessary at all given that `DELETE` method is the least likely to go through various firewalls.
[swagger.txt](https://github.com/wultra/powerauth-webflow/files/6337599/swagger.txt)

petrdvorak (approved):
Looks OK

PR #988, opened 2021-04-23
Fix #987: Use null value in NO_USERNAME algorithm

### LINKED ISSUE #987: Use null value in NO_USERNAME algorithm
An error occurs due to index violation for credentials with no username at all:

 ```
"Unknown error occurred, cause: could not execute statement; SQL [n/a]; constraint [POWERAUTH_WEBFLOW.NS_CREDENTIAL_STORAGE_QUERY1]; 
nested exception is org.hibernate.exception.ConstraintViolationException: could not execute statement" 
```

Instead of using empty string for username, the username needs to be persisted as `NULL` to avoid duplicated in database.

romanstrobl:
The `null` value for username needs to be used for credentials which have no usernames at all due to unique index constraint.

In Oracle multiple `null` values are not allowed in composite unique indexes. Thus, the unique index definition has to handle the `null` username case implicitly. A non-unique index for username was added for Oracle for performance reasons, too. 

In other supported databases multiple `null` values are allowed in composite unique indexes, so no DDL changes are required and the database queries correctly process both use cases (`null` usernames for PINs and `non-null` unique usernames for passwords).


petrdvorak (approved):
Looks OK

PR #989, opened 2021-04-23
Fix #977: Track last username change timestamp

### LINKED ISSUE #977: Track last username change timestamp
We should track the timestamp of last username change and expose it similar to last credential change timestamp.

romanstrobl:


powerauth-nextstep/src/main/java/io/getlime/security/powerauth/app/nextstep/service/CredentialService.java
```diff
@@ -625,6 +626,7 @@ public CredentialSecretDetail createCredential(UserIdentityEntity user, Credenti
         credential.setEncryptionAlgorithm(protectedCredentialValue.getEncryptionAlgorithm());
         credential.setHashingConfig(credentialDefinition.getHashingConfig());
         credential.setTimestampLastCredentialChange(new Date());
+        credential.setTimestampLastUsernameChange(new Date());
```
petrdvorak, line 629: Store the `new Date()` in variable to avoid having different values for the two timestamps.

petrdvorak (changes requested):
The duplicate `new Date()` in the method parameter is not optimal, it will result in different timestamps for events that occurred in the same call...

petrdvorak (approved):
Looks OK

PR #990, opened 2021-04-26
Fix #985: Consider using Spring Data JPA Streams for queries with many results

### LINKED ISSUE #985: Consider using Spring Data JPA Streams for queries with many results
Currently, we sometimes have calls like this in the code:

```java
credentialRepository.findAllByStatus(CredentialStatus.ACTIVE);
```

This will make JVM explode since all results will be fetched and in memory at once. Paging would help, but why suffer when there are Streams...

https://www.baeldung.com/spring-data-java-8

romanstrobl:


powerauth-nextstep/src/main/java/io/getlime/security/powerauth/app/nextstep/service/UserIdentityService.java
```diff
@@ -642,13 +644,26 @@ private void removeAllCredentials(UserIdentityEntity user) {
      * @param user User identity entity.
      */
     private void removeAllOtps(UserIdentityEntity user) {
-        final List<OtpEntity> otps = otpRepository.findAllByUserId(user.getUserId());
-        otps.forEach(otp -> {
-            if (otp.getStatus() != OtpStatus.REMOVED) {
+        List<OtpEntity> removedOtps = new ArrayList<>();
+        try (final Stream<OtpEntity> activeOtps = otpRepository.findAllByUserIdAndStatus(user.getUserId(), OtpStatus.ACTIVE)) {
```
petrdvorak, line 648: Maybe again a task for the Update query instead of the Select via Stream?

petrdvorak (approved):
Looks OK, I have one optimization tip - maybe we could create a query to update the items directly, instead of selecting them and saving the collection?

powerauth-nextstep/src/main/java/io/getlime/security/powerauth/app/nextstep/service/UserIdentityService.java
romanstrobl: Ah, yes, that would be better.

PR #993, opened 2021-04-29
Fix #967: Improve logging in Next Step

### LINKED ISSUE #967: Improve logging in Next Step
Currently, Next Step logs all events during requests, responses and for various errors. 

Let's improve the logging a bit more:
- Next Step client calls from Web Flow should log Next Step error details (currently only HTTP statuses are logged).
- Log additional details in Next Step services.
- Add debug logging for important parameter values and for status changes.

romanstrobl:
Additional logging:
- error codes from the REST API responses are logged by clients whenever they are available instead of generic HTTP errors
- additional INFO logs during authentication and for counter reset
- many new DEBUG messages throughout whole Next Step for easier debugging


petrdvorak (approved):
Looks OK

PR #995, opened 2021-04-29
Fix #994: Allow setting custom credential expiration for temporary credential migration

### LINKED ISSUE #994: Allow setting custom credential expiration for temporary credential migration
The credential expiration for temporary credentials is set using policy configuration: `temporaryCredentialExpirationTime`. This works well for new credentials, however it does not allow exact configuration for imported credentials during migration.  For this purpose the `expirationTimestamp` should be added as a parameter for created and updated credentials.

romanstrobl:
I added the possibility to override credential expiration timestamp into all cases where credential expiration is handled.


petrdvorak (approved):
Looks OK

PR #1000, opened 2021-05-07
Fix #997: Add option to disallow login with temporary credentials in Web Flow

### LINKED ISSUE #997: Add option to disallow login with temporary credentials in Web Flow
For the PSD2 use case, Web Flow should disallow login with temporary credentials because Web Flow does not have a change password form, so the credential cannot be changed to become permanent by Web Flow. We should add a configuration option to enable or disable this behavior for temporary credentials because the use case for temporary credentials might differ in various deployments.

So, the Web Flow UI will use a fake SMS and password authentication page for the case when the credential is temporary, like in the case of an unknown username.

romanstrobl:
This pull request implements credential type check. By default it is not possible to login using TEMPORARY credentials but it can be changed in configuration.

The implementation differs slightly for 2 cases:
- username and password authentication: authentication fails for TEMPORARY credentials
- SCA use case: fake SMS authentication page is presented to user for TEMPORARY credentials

In case Data Adapter proxy is enabled in Next Step the information about credential type is not available, the behaviour is the same as before, it is expected that system which performs the authentication does not allow it for TEMPORARY credentials if required.


petrdvorak (approved):
Looks OK

PR #1011, opened 2021-06-01
Fix #1009: Include Bouncy Castle library in applications

### LINKED ISSUE #1009: Include Bouncy Castle library in applications
Now that PowerAuth server includes Bouncy Castle library in the war file (https://github.com/wultra/powerauth-server/pull/583), we should also include the library in Web Flow and Next Step war files.

romanstrobl:


petrdvorak (approved):
Looks OK

PR #1013, opened 2021-06-02
Fix #1012: Cors exception in Web Flow

### LINKED ISSUE #1012: Cors exception in Web Flow
Exception seems to occur since last dependency update:
```
java.lang.IllegalArgumentException: When allowCredentials is true, allowedOrigins cannot contain the special value "*" since that cannot be set on the "Access-Control-Allow-Origin" response header. To allow credentials to a set of origins, list them explicitly or consider using "allowedOriginPatterns" instead.
```

romanstrobl:
Spring Framework introduced a new method for Cors patterns: `setAllowedOriginPatterns`.


petrdvorak (approved):
Looks OK

PR #1021, opened 2021-06-14
Fix #1020: Document Bouncy Castle deployment changes

### LINKED ISSUE #1020: Document Bouncy Castle deployment changes
The Bouncy Castle library is now included with applications. We should document this in the migration guide.

romanstrobl:


petrdvorak (approved):
Looks OK

PR #1030, opened 2021-06-16
Fix #1029: Avoid using double quotes in DDL, unify naming conventions in DDL

### LINKED ISSUE #1029: Avoid using double quotes in DDL
Double quotes imply case sensitivity in Oracle and PostgreSQL, avoid using them in DDL.

romanstrobl:


petrdvorak (commented):
Do we need to provide the alter script for this for existing databases? 🤔

romanstrobl:
I do not think so, because current DDL works fine with JPA, this is just a cleanup for the next release.

petrdvorak (approved):
Approved, after offline chat.

PR #1032, opened 2021-06-17
Fix #1031: OTP authentication in checkOnly mode

### LINKED ISSUE #1031: OTP authentication in checkOnly mode
The `/auth/otp` endpoint should be also available in `checkOnly` mode when the OTP value is checked, however the OTP status change is not persisted for successful authentication. This mode is used when credentials are stored in an external system and the result of OTP check is used as an input to the authentication using the external system.

The following changes are required for implementing this scenario:
- new `checkOnly` parameter in `/auth/otp` request object
- updated Next Step client methods
- OTP status is not changed due to successful verification in `checkOnly` mode
- OTP failed attempt counter is still updated even for `checkOnly` mode and OTP is blocked when configured number of failed attempts is exceeded to prevent brute force attacks
- OTP time limit is still checked even for `checkOnly` mode to prevent time related attacks
- parameter `updateOperation` should be set to `false` for `checkOnly` mode (operation is not updated)
- credential counter is not updated because credential is stored in an external system
- separate logging and auditing for the `checkOnly` mode
- tests for the `checkOnly` mode

It is expected that the `/auth/otp` endpoint is called once again after successful authentication against external system, so that the OTP status is set to `USED`. In case the OTP is created within an operation, `updateOperation=true` parameter should be used to update the operation, too.

romanstrobl:
For details of this change, see: https://github.com/wultra/powerauth-webflow/issues/1031

### DIFF
--- a/powerauth-nextstep/src/main/java/io/getlime/security/powerauth/app/nextstep/service/AuthenticationService.java
+++ b/powerauth-nextstep/src/main/java/io/getlime/security/powerauth/app/nextstep/service/AuthenticationService.java
@@ -427,10 +427,8 @@ public class AuthenticationService {
             authenticationRepository.save(authentication);
         }
 
-        String mode = "authentication";
-        if (checkOnly) {
-            mode = "check";
-        }
+        final String mode = checkOnly ? "check" : "authentication";
+
         logger.info("OTP " + mode + " result: {}, remaining attempts: {}, user ID: {}, user identity status: {}, OTP status: {}, credential status: {}, operation failed: {}",
                 authenticationResult, remainingAttempts, user == null ? null : user.getUserId(), user == null ? null : user.getStatus(), otp.getStatus(), credential == null ? null : credential.getStatus(), operationFailed);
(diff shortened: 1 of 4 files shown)


powerauth-nextstep/src/main/java/io/getlime/security/powerauth/app/nextstep/service/AuthenticationService.java
```diff
@@ -423,7 +427,11 @@ public OtpAuthenticationResponse authenticateWithOtp(OtpAuthenticationRequest re
             authenticationRepository.save(authentication);
         }
 
-        logger.info("OTP authentication result: {}, remaining attempts: {}, user ID: {}, user identity status: {}, OTP status: {}, credential status: {}, operation failed: {}",
+        String mode = "authentication";
```
petrdvorak, line 430: ... be a bit fancy:

```java
final String mode = checkOnly ? "check" : "authentication";
```

petrdvorak (approved):
Looks good, I decided to comment on a figure skating judges level... ;-)

PR #1035, opened 2021-06-21
Fix #1034: Error while generating Javadoc

### LINKED ISSUE #1034: Error while generating Javadoc
Error:
```
org.apache.maven.reporting.MavenReportException: 
Exit code: 2 - error: option --add-exports not allowed with target 8

Command line was: /Library/Java/JavaVirtualMachines/zulu-15.jdk/Contents/Home/bin/javadoc @options @packages

Refer to the generated Javadoc files in '/Users/romanstrobl/Wultra/powerauth-webflow/powerauth-tpp-engine-model/target/apidocs' dir.
```

romanstrobl:
I suggest we switch the source and target to Java 11. This will not only fix the Javadoc errors, but it will cause a migration to a new and better supported LTS version of Java.

Java 8 has gone through the End of Public Updates process for legacy releases, although commerical support is still available.

### DIFF
--- a/docs/Web-Flow-1.1.0.md
+++ b/docs/Web-Flow-1.1.0.md
@@ -1,5 +1,9 @@
 # Migration from 1.0.0 to 1.1.0
 
+## Updated Minimum Java Runtime Version
+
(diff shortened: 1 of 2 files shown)


hvge (approved):
Looks OK

PR #1043, opened 2021-08-17
Fix #1042: Invalid timestamp definition without default value for MySQL

### LINKED ISSUE #1042: Invalid timestamp definition without default value for MySQL
Some versions of MySQL do not accept timestamp definition without default value. The `NULL` value needs to be set explicitly for this case.

romanstrobl:


petrdvorak (approved):
Looks OK

PR #1057, opened 2021-11-15
Fix #1056: Do not apply user identity foreign keys by default

### LINKED ISSUE #1056: Do not apply user identity foreign keys by default
Update the DDL for Next Step, so that user identity foreign keys are not applied by default, which breaks the end-to-end tests.

romanstrobl:


saalistaja (approved):
Looks good.

PR #1070, opened 2022-01-03
Fix #1069: Missing @NoArgsConstructor annotations in model

### LINKED ISSUE #1069: Missing @NoArgsConstructor annotations in model
The `@NoArgsConstructor` annotation is missing for several model classes in Next Step which causes following error:

```
Cannot construct instance of `io.getlime.security.powerauth.lib.nextstep.model.entity.BankAccountDetail` (no Creators, like default constructor, exist): cannot deserialize from Object value (no delegate- or property-based Creator)
```

romanstrobl:
Add missing `@NoArgsConstructor` annotations in model which solve serialization issues.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -8,7 +8,7 @@
 
     <groupId>io.getlime.security</groupId>
     <artifactId>powerauth-webflow-parent</artifactId>
-    <version>1.2.0</version>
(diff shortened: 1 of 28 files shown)


petrdvorak (approved):
Looks OK

PR #1081, opened 2022-01-13
Fix #1080: Avoid canceling operation when WebView sleeps

### LINKED ISSUE #1080: Avoid canceling operation when WebView sleeps
When the user switches to another application on iOS, the WebView gets hibernated after a short while. This causes a websocket `SessionDisconnectEvent` on the server with abnormal CloseStatus `NO_CLOSE_FRAME`. We should avoid canceling the operation in case such event occurs because the application can be reopened and the flow needs to continue even with a closed websocket connection. 

On desktop the CloseStatus `NORMAL` is sent when the browser is closed.

Ideally we should prevent the WebView from going to sleep completely. However, it seems that there is no easy way to make this happen, e.g. a video is played to avoid the WebView sleeping or other similar workaround is implemented to achieve this.

romanstrobl:


petrdvorak (approved):
Looks great, thank you for identifying this!

PR #1125, opened 2022-11-29
Allow targeting to JDK 8

romanstrobl:
Support for JDK 8, thanks to @saalistaja.

I suggest we merge it to `1123-payment-signing` branch, unfortunately there are still clients who insist on JDK 8.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -89,8 +89,9 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <maven.compiler.source>11</maven.compiler.source>
-        <maven.compiler.target>11</maven.compiler.target>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -89,8 +89,9 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <maven.compiler.source>11</maven.compiler.source>
-        <maven.compiler.target>11</maven.compiler.target>
+<!--        <maven.compiler.source>11</maven.compiler.source>-->
+<!--        <maven.compiler.target>8</maven.compiler.target>-->
```
banterCZ, line 93: What about to keep this?
```suggestion
        <maven.compiler.target>8</maven.compiler.target>
```

banterCZ (approved):

PR #1128, opened 2022-11-30
Add github action for JDK 8

banterCZ:


banterCZ:
Not needed anymore after #1130

PR #1130, opened 2022-11-30
Kill sun.security imports

romanstrobl:
This should help with compilation issues on various JDK versions.


petrdvorak (approved):
Looks OK, thank you for the improvement.

PR #1135, opened 2022-11-30
Issues/1134 exclude tomcat

banterCZ:


petrdvorak (approved):
Looks OK, we can polish this later if needed.

romanstrobl:
👍

PR #1144, opened 2022-12-01
Fix #1143: Cleanup compiler configuration

### LINKED ISSUE #1143: Cleanup compiler configuration

banterCZ:


romanstrobl (changes requested):
We use source level 8 for this release, all code was just migrated to it. Let's postpone this change after release if you do not mind. You can leave the PR open, I suggest to merge it after release, because the current configuration is tested and there is no need to migrate to Java 11 yet.

banterCZ:
> We use source level 8 for this release, all code was just migrated to it. Let's postpone this change after release if you do not mind. You can leave the PR open, I suggest to merge it after release, because the current configuration is tested and there is no need to migrate to Java 11 yet.

I agree, it does not have to be merged now. Just a prepared cleanup.

petrdvorak (approved):
Looks OK

banterCZ:
Obsolete, since we would like to upgrade to Spring 6 and JDK 17.

PR #1146, opened 2022-12-01
Fix #1145: Simplify exclusion of tomcat dependency

### LINKED ISSUE #1145: Simplify exclusion of tomcat dependency

banterCZ:
Just a clean up, may be postponed after release.


romanstrobl (approved):
Looks OK.

PR #1156, opened 2022-12-02
Fix #1155: Improve TppEngineClient error handling

### LINKED ISSUE #1155: Improve TppEngineClient error handling
`TppEngineClient` should return error response body.

banterCZ:


petrdvorak (approved):
Looks OK

powerauth-tpp-engine-client/src/main/java/io/getlime/security/powerauth/lib/tpp/engine/client/TppEngineClient.java
```diff
@@ -373,13 +375,33 @@ private void logError(TppEngineClientException ex) {
...
+
+        final Error error = e.getErrorResponse().getResponseObject();
+        return new TppEngineError(error.getCode(), error.getMessage());
+    }
+
     /**
      * Resolve error code based on HTTP status code from REST client exception.
      */
-    private String resolveErrorCode(RestClientException ex) {
+    private static String resolveErrorCode(RestClientException ex) {
```
romanstrobl, line 401: Just out of curiosity, is there a reason for the `static` modifier?

romanstrobl: But all code runs inside of an instance, so what is the point? It doesn't matter, just it is a bit misleading to me, but OK.

banterCZ: IMHO once the method has been made static, it is clear that it can be moved out of the class.

PR #1161, opened 2022-12-04
Fix #1139: Allow selection of certificate for signer during operation approval

### LINKED ISSUE #1139: Allow selection of certificate for signer during operation approval
Currently only the first certificate which matches filter `ICAClientSign.CERTLOAD_SIGNING_FLAG | ICAClientSign.CERTLOAD_QUALIFIED_FLAG` is extracted from the keystore. We want to allow users select a certificate based on following values:
- issuer
- expiration date
- subject name
- serial number

The select box for certificate selection should be a new React component which allows the selection with rich data displayed from parsed certificate PEM.

romanstrobl:


petrdvorak (approved):
Looks OK

banterCZ (approved):
Looks good.

PR #1162, opened 2022-12-05
Fix #1141: Add support for external operation data for signing using certificate

### LINKED ISSUE #1141: Add support for external operation data for signing using certificate
Currently, `operation_data` is used for singing with certificate during operation approval.

We should extend the Data Adapter `initAuthMethod` method so that it can be used for retrieving external operation data.

The external operation data should be sent to React frontend for signing. The signed message is verified using existing `verifyCertificate` method.

romanstrobl:


powerauth-webflow/src/main/java/io/getlime/security/powerauth/app/webflow/controller/HomeController.java
```diff
@@ -388,5 +388,6 @@ private void cleanHttpSession() {
         httpSession.removeAttribute(HttpSessionAttributeNames.USERNAME);
         httpSession.removeAttribute(HttpSessionAttributeNames.CLIENT_CERTIFICATE);
         httpSession.removeAttribute(HttpSessionAttributeNames.APPROVAL_BY_CERTIFICATE_ENABLED);
+        httpSession.removeAttribute(HttpSessionAttributeNames.OPERATION_DATA_EXTERNAL);
```
banterCZ, line 391: This is not synchronized. Does it mind?

romanstrobl: I will add synchronization.a

PR #1168, opened 2022-12-06
Fix #1167: TPP Engine client uses PUT call instead of POST

### LINKED ISSUE #1167: TPP Engine client uses PUT call instead of POST
This is a regression from rewriting to a new REST client.

petrdvorak:


romanstrobl (approved):
Looks OK.

PR #1178, opened 2022-12-08
Fix #1177: OAuth2 refresh token is not issued

### LINKED ISSUE #1177: OAuth2 refresh token is not issued
Relates to #1169

banterCZ:


romanstrobl (approved):
Looks OK.

PR #1186, opened 2022-12-12
Fix #1185: Remove hsqldb dependency

### LINKED ISSUE #1185: Remove hsqldb dependency

banterCZ:


powerauth-webflow-client/pom.xml
```diff
@@ -86,13 +86,6 @@
             <artifactId>powerauth-nextstep-client</artifactId>
         </dependency>
 
-        <!-- H2 DB -->
-        <dependency>
-            <groupId>org.hsqldb</groupId>
-            <artifactId>hsqldb</artifactId>
```
banterCZ, line 92: hsqldb is not h2, if needed for the demo in the future, should be enabled only for a specific profile.

romanstrobl (approved):
I have verified that the client still works.

PR #1188, opened 2022-12-12
workflow for Coverity scan

zcgandcomp:


banterCZ (approved):
The commit is not signed, otherwise very good.

PR #1192, opened 2022-12-15
Fix #1191: OAuth2 refresh token is not issued

### LINKED ISSUE #1191: OAuth2 refresh token is not issued
Follow-up to #1177

banterCZ:
- set `refresh_token` to authorized grant type


romanstrobl (approved):
This should help.

petrdvorak (approved):
looks ok

PR #1194, opened 2022-12-16
Fix #1193: Coverity - Dereference null return value at ICACertificateParser

### LINKED ISSUE #1193: Coverity - Dereference null return value at ICACertificateParser

banterCZ:


romanstrobl (approved):
Looks good.

PR #1215, opened 2022-12-23
Fix #1214: Standalone dev run

### LINKED ISSUE #1214: Standalone dev run
Make possible to run dev environment as standalone.

banterCZ:


romanstrobl (changes requested):
We should include the Next Step and WebFlow client applications, too.

powerauth-webflow-client/pom.xml
```diff
@@ -148,6 +148,15 @@
                 </plugins>
             </build>
         </profile>
+        <profile>
```
romanstrobl, line 151: The client does not require connection to Postgres as far as I know.

romanstrobl (approved):
Approved, however I would suggest to remove database-related configuration for powerauth-webflow-client, it should not be required.

powerauth-webflow-client/pom.xml
banterCZ: Good catch, I have change that to H2.

PR #1223, opened 2023-01-05
Fix #1222: Redesign token revoke endpoint

### LINKED ISSUE #1222: Redesign token revoke endpoint
The token revoke endpoint should:
- use the `POST` HTTP method so that request body can always be sent, the `DELETE` method is problematic in some environments
- correctly authenticate users using HTTP basic authentication (using `client_id` and `client_secret`)
- be implemented according to RFC 7009: https://www.rfc-editor.org/rfc/rfc7009

romanstrobl:


powerauth-webflow/src/main/java/io/getlime/security/powerauth/app/webflow/controller/TokenRevocationController.java
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * Endpoint used for token revocation, see: https://www.rfc-editor.org/rfc/rfc7009
+     * @param user Authenticated user.
+     * @param token OAuth 2.0 access token or refresh token for revocation.
+     * @param tokenTypeHint Optional hint whether token is an access token or a refresh token.
+     * @throws InvalidTokenException Thrown when token is invalid.
+     */
+    @PostMapping("/token/revoke")
+    @ResponseBody
+    public void revokeToken(@AuthenticationPrincipal User user, @RequestParam("token") String token, @RequestParam("token_type_hint") String tokenTypeHint) throws InvalidTokenException {
```
banterCZ, line 65: If `tokenTypeHint` is optional, I would expect:
```suggestion
    public void revokeToken(@AuthenticationPrincipal User user, @RequestParam("token") String token, @RequestParam(name = "token_type_hint", required = false) String tokenTypeHint) throws InvalidTokenException {
```

powerauth-webflow/src/main/java/io/getlime/security/powerauth/app/webflow/exception/DefaultExceptionResolver.java
```diff
@@ -120,4 +119,19 @@ public String handleUnauthorizedException(InsufficientAuthenticationException ex
...
+    /**
+     * Handling of invalid token exception.
+     * @param ex Exception.
+     * @return Response with error details.
+     */
+    @ExceptionHandler(InvalidTokenException.class)
+    @ResponseStatus(HttpStatus.BAD_REQUEST)
+    public @ResponseBody ErrorResponse handleInvalidTokenException(InvalidTokenException ex) {
+        logger.warn("Error occurred in Web Flow server: {}", ex.getMessage());
+        audit.warn("Error occurred in Web Flow server", AUDIT_DETAIL_BAD_REQUEST, ex);
```
banterCZ, line 132: I see the whole exception in audit, but I am thinking about logging it at `debug` level too.
```suggestion
        logger.warn("Error occurred in Web Flow server: {}", ex.getMessage());
        logger.debug("Error occurred in Web Flow server", ex);
        audit.warn("Error occurred in Web Flow server", AUDIT_DETAIL_BAD_REQUEST, ex);
```

powerauth-webflow/src/main/java/io/getlime/security/powerauth/app/webflow/service/TokenRevocationAuthenticationService.java
```diff
@@ -0,0 +1,63 @@
...
+            return clientId;
+        }
+        throw new InsufficientAuthenticationException("Invalid authentication during token revocation");
+    }
+
+    /**
+     * @param userDetails User details.
+     * @return OAuth 2.0 client ID.
+     */
+    protected String getClientId(UserDetails userDetails) {
```
banterCZ, line 59: Do we need `protected` access modifier?

banterCZ (approved):
Looks good.

powerauth-webflow/src/main/java/io/getlime/security/powerauth/app/webflow/controller/TokenRevocationController.java
romanstrobl: Right, I will make the hint optional.

powerauth-webflow/src/main/java/io/getlime/security/powerauth/app/webflow/exception/DefaultExceptionResolver.java
romanstrobl: We'd have to change this in all exception handlers for consistency, let's handle such changes separately.

powerauth-webflow/src/main/java/io/getlime/security/powerauth/app/webflow/service/TokenRevocationAuthenticationService.java
romanstrobl: Typo, it should be `private`.

petrdvorak (approved):
Looks OK

PR #1225, opened 2023-01-16
Fix #1224: Update token revoke endpoint configuration

### LINKED ISSUE #1224: Update token revoke endpoint configuration
Reflect the OAuth token revoke endpoint change in Spring Security.

romanstrobl:


petrdvorak (approved):
Looks good

PR #1247, opened 2023-03-30
Fix #1246: Turn off auto-commit

### LINKED ISSUE #1246: Turn off auto-commit
See https://vladmihalcea.com/why-you-should-always-use-hibernate-connection-provider_disables_autocommit-for-resource-local-jpa-transactions/

banterCZ:


romanstrobl (approved):
Looks OK.

PR #1289, opened 2023-05-11
Fix #1284: Review Java warnings

### LINKED ISSUE #1284: Review Java warnings
Review Java warnings specifically after upgrading to JDK 17.

banterCZ:
- Remove redundant throws clause
- Add final modifier
- Use @serial annotation
- Use getSubjectX500Principal instead of deprecated getSubjectDN
- Replace deprecated resolver.setCookieName() by constructor call
- Remove unnecessary return as the last statement
- Replace .collect(Collectors.toList()) by toList()
- Use standard charsets constants
- Micro-optimalisation of containsAll
- Make inner class static
- Simplify optional call chain
- Replace explicit type by diamond operator <>
- Use List.sort instead of Collections.sort
- Use expression lambda instead of statement
- Use List.of
- Use text block
- Use patter variable
- Enhance switch
- Convert AuthorizationCode, OfflineSignatureQrCode, Violation and CertInfo to record


romanstrobl (approved):
Thank you for the cleanup after migration to Java 17!

PR #1301, opened 2023-05-16
Fix #1300: Coverity: Dereference null return value AuthenticationService

### LINKED ISSUE #1300: Coverity: Dereference null return value AuthenticationService

banterCZ:


romanstrobl (approved):
Good catch!

PR #1306, opened 2023-05-19
Fix #1304: Update documentation for soft and hard limits in credential policy

### LINKED ISSUE #1304: Update documentation for soft and hard limits in credential policy
The value `null` can be used for disabling soft and hard limits, not `0`.

romanstrobl:


petrdvorak (approved):
Looks OK

PR #1317, opened 2023-05-29
New text elements for WebFlow.

zcgandcomp:


powerauth-webflow/package.json
```diff
@@ -19,43 +19,44 @@
...
+    "axios": "^0.27.2",
+    "bluebird": "^3.7.2",
+    "intl": "^1.2.5",
+    "lodash": "^4.17.21",
     "react": "^16.14.0",
     "react-bootstrap": "^0.33.1",
-    "react-select": "1.3.0",
+    "react-device-detect": "^2.2.2",
     "react-dom": "^16.14.0",
+    "react-flags-select": "^2.2.3",
```
romanstrobl, line 31: Let's revert all changes except for `react-flags-select`, because many of the dependencies are a downgrade.

romanstrobl (changes requested):
Looks OK, let's just revert the dependency changes, we just need to add 1 new dependency and leave the rest as they are.

romanstrobl (approved):
Looks OK.

powerauth-webflow/package.json
zcgandcomp: package.json updated and currently contains no downgrades.

romanstrobl (approved):

PR #1327, opened 2023-05-31
Fix #1324: Allow case insensitive username comparison

### LINKED ISSUE #1324: Allow case insensitive username comparison
Currently, the usernames are case-sensitive in our system. This is typically not desired. We should add a flag to the credential definition to indicate if usernames should be case-sensitive and use case-insensitive comparison as default (as this will typically be the case).

petrdvorak:


zcgandcomp (approved):
- We change the create and update. 
- Lookup is also case-insensitive. 
- Generation of username also checks the generated username case-insensitive. 

Looks Good.

PR #1338, opened 2023-06-01
Fix #1337: Add favicon.ico to the accessible resources

### LINKED ISSUE #1337: Add favicon.ico to the accessible resources
Web browser automatically hits `https://${BASEURL}/favicon.ico` on a page load. If our application runs in the root context, this call will now trigger 401 in our application and initiate internal resource redirect of the icon file to `/authenticate` endpoint, which will drop our session at random stages of the flow...

petrdvorak:


romanstrobl (approved):
Looks OK.

PR #1374, opened 2023-07-04
Fix #1367: Invalid constraint for audit_type

### LINKED ISSUE #1367: Invalid constraint for audit_type
The `audit_type` is set to `NOT NULL` in Liquibase scripts, which does not correspond to:
https://github.com/wultra/lime-java-core/blob/develop/docs/sql/postgresql/create_schema.sql#L8

jandusil:
- Remove the NOT NULL constraint in the liquibase changeset


romanstrobl (approved):
Looks OK.

PR #1381, opened 2023-07-28
Fix #1380: Make versions of maven plugins implicit

### LINKED ISSUE #1380: Make versions of maven plugins implicit
Make versions of maven plugins implicit and dependent on Spring Boot.

banterCZ:


romanstrobl (approved):
Nice cleanup!

PR #1387, opened 2023-07-28
Fix #1323: NULL the username when removing the user / user credentials

### LINKED ISSUE #1323: NULL the username when removing the user  / user credentials
When a user or his credentials are removed, his username in credentials remains in the database with the status `REMOVED`. This prevents to create a new user with the same username. 

Set username to NULL when the user or his credentials are removed (the original value of the username should be retained in audit and credential history)

banterCZ:


romanstrobl (approved):
Thanks for fixing this issue.

PR #1389, opened 2023-07-28
Fix #1388: Hibernate no longer supports disabling the use of new ID generator mappings

### LINKED ISSUE #1388: Hibernate no longer supports disabling the use of new ID generator mappings

banterCZ:


romanstrobl (approved):
Looks OK.

PR #1395, opened 2023-08-07
Fix #1394: Error while loading lang.json

### LINKED ISSUE #1394: Error while loading lang.json
The language setting is not read correctly in the vanilla version. 

` Caused by: java.net.MalformedURLException: no protocol: lang.json
`

zcgandcomp:


powerauth-webflow/src/main/java/io/getlime/security/powerauth/app/webflow/configuration/WebMvcConfiguration.java
```diff
@@ -57,6 +57,9 @@ public class WebMvcConfiguration implements WebMvcConfigurer {
     @Autowired
     private WebFlowServerConfiguration configuration;
 
+    @Autowired
+    ResourceLoader resourceLoader;
```
banterCZ, line 61:
```suggestion
    private ResourceLoader resourceLoader;
```

banterCZ (approved):
Looks good.

PR #1399, opened 2023-08-14
Fix #1397: Update documentation for PUT /credential

### LINKED ISSUE #1397: Update documentation for PUT /credential
Add attribute `credentialValue` to the request object in the dev portal documentation.

zcgandcomp:


docs/Next-Step-Server-REST-API-Reference.md
```diff
@@ -7973,6 +7973,7 @@ The list of expected status codes:
     "credentialName": "RETAIL_CREDENTIAL",
     "credentialType": "PERMANENT",
     "username": "username1234",
+    "credentialValue": "JeM1vr%GyJFh",
```
jandusil, line 7976: Isn't missing also timestampExpires property from UpdateCredentialRequest?

zcgandcomp: Yes, it is. However, we should describe in the future which parameter is optional and which is mandatory - for the entire API documentation.

PR #1420, opened 2023-09-14
Fix #1419: Redundant call of Duration.ofSeconds

### LINKED ISSUE #1419: Redundant call of Duration.ofSeconds

banterCZ:


romanstrobl (approved):
Looks OK.

PR #1421, opened 2023-09-14
Fix #1412: Review liquibase scripts

### LINKED ISSUE #1412: Review liquibase scripts

jandusil:
- Review sequences


romanstrobl (changes requested):
This will probably end up with many errors like this:

```
ERROR: value "9999999999999999999999999999" is out of range for type bigint
```

romanstrobl (changes requested):
The `minValue` setting seems to be inconsistent.

docs/db/changelog/changesets/powerauth-nextstep/1.4.x/20230324-init-db.xml
```diff
@@ -10,7 +10,7 @@
             </not>
         </preConditions>
         <comment>Create a new sequence ns_operation_afs_seq</comment>
-        <createSequence sequenceName="ns_operation_afs_seq" cacheSize="20" />
+        <createSequence sequenceName="ns_operation_afs_seq" startValue="1" incrementBy="1" cacheSize="20" minValue="1" />
```
romanstrobl, line 13: We set `minValue` here however not in other sections. We should probably always se this.

romanstrobl: Ah, I see there is already the `startValue`, so this setting seems rather redundant. Anyway it is not consistent.

jandusil: Apologies, this was forgotten by mistake

zcgandcomp (approved):
Looks OK.

romanstrobl (approved):
Looks good now.

PR #1424, opened 2023-09-18
Fix #1363: The entered credentials are removed upon submit

### LINKED ISSUE #1363: The entered credentials are removed upon submit
The user enters his credentials (password and OTP)  and press confirm

Expected behavior:
Show loading spinner
Show either a successful result or a reseted form and error message.
 
Actual behavior:
The password and OTP are immediately removed.

Note 
In smsComponent.js the `handleSubmit` function set empty values of password and OTP into the state. This should be done after the `smsAuthAction.authenticate` dispatch an error

banterCZ:


romanstrobl (approved):
Looks OK.

PR #1428, opened 2023-09-20
Fix #1427: Restrict import of javax.annotation

### LINKED ISSUE #1427: Restrict import of javax.annotation
Spring 6 migrated to `jakarta.annotations` but `guava` depends on `jsr305` and still transitively brings `javax.annotation.*`. Let's keep the dependency but not use the `javax` imports in our code.

banterCZ:


romanstrobl (approved):
Looks OK.

PR #1435, opened 2023-10-03
Fix #1434: Add logstash-logback-encoder dependency

### LINKED ISSUE #1434: Add logstash-logback-encoder dependency

banterCZ:


romanstrobl (approved):
Looks OK.

PR #1466, opened 2023-10-31
Fix #1462: Update documentation regarding AS support

### LINKED ISSUE #1462: Update documentation regarding AS support
The `Web-Flow-Installation-Manual.md` contains download link to Tomcat 9. Change it to link Tomcat 10.

jnpsk:
Fix #1462
In the manual, there is download link to Tomcat 9. The link was changed to point to Tomcat 10.


banterCZ (approved):
Perfect! This should be also backported to 1.5.x

PR #1517, opened 2023-12-20
Fix #1513: The request to non existent resource ends in 302

### LINKED ISSUE #1513: The request to non existent resource ends in 302
Request resource that doesn't exist but is permitted without authorization- e.g., `/resources/nonexistent.html`.

Expected result:
HTTP code 404

Actual result:
HTTP code 302 to location /authenticate

banterCZ:


zcgandcomp (approved):
Looks OK.

Tested successfully.

PR #1519, opened 2023-12-20
Fix #1518: Prefer List.of over guava ImmutableList

### LINKED ISSUE #1518: Prefer List.of over guava ImmutableList

banterCZ:


jnpsk (approved):
nice

PR #1523, opened 2023-12-20
Fix #1520: Allowed languages are not working

### LINKED ISSUE #1520: Allowed languages are not working
Default languages should be explicitly stated in the reducer and the index.html template otherwise, the language switch will not work correctly.

zcgandcomp:


banterCZ (approved):
Looks good.

PR #1524, opened 2023-12-21
Issues/1521 language support backport

zcgandcomp:


romanstrobl (approved):
Looks OK. I think we are getting to a point where it would be nice to introduce a generic language support with configuration of supported languages.

PR #1537, opened 2024-01-05
Fix #1535: Remove Guava dependency

### LINKED ISSUE #1535: Remove Guava dependency
Prefer JDK over Guava API. 

Guava still used in `WebSocketHandshakeInterceptor` for validation IP adresses, see alternatives https://www.baeldung.com/java-validate-ipv4-address

banterCZ:
Guava dependency remains because of owasp-java-html-sanitizer, but the API not directly used anymore.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -257,6 +257,19 @@
                 <version>${bcprov-jdk18on.version}</version>
             </dependency>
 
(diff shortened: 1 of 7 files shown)


romanstrobl (approved):
Looks OK.

PR #1557, opened 2024-01-25
Fix #1555: create_schema.sql in 1.6.0 branch is not valid

### LINKED ISSUE #1555: create_schema.sql in 1.6.0 branch is not valid
SQL Script https://github.com/wultra/powerauth-webflow/blob/releases/1.6.x/docs/sql/postgresql/create_schema.sql fails with error
```
ERROR:  there is no unique constraint matching given keys for referenced table "oauth2_registered_client"
```
check other db's for the same

zcgandcomp:
(cherry picked from commit faf65bdb012af5055b998eeb974d6a1c65edb0f5)

### DIFF
--- a/docs/sql/oracle/create_schema.sql
+++ b/docs/sql/oracle/create_schema.sql
@@ -76,6 +76,8 @@ CREATE TABLE oauth2_registered_client (
     PRIMARY KEY (id)
 );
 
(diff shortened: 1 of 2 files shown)


korbelm (approved):

banterCZ (approved):
Thanks for the backport.

PR #1570, opened 2024-02-07
Fix #1568: NextStep OpenID Connect (OIDC) Support

### LINKED ISSUE #1568: NextStep OpenID Connect (OIDC) Support
Support authentication via OpenID Connect (OIDC). Because of the backward compatibility, there should be no authentication by default.

banterCZ:


romanstrobl (approved):
Looks OK.

powerauth-nextstep/src/main/resources/application.properties
```diff
@@ -70,5 +70,15 @@ banner.application.version=@project.version@
 
 logging.config=${POWERAUTH_NEXTSTEP_LOGGING:}
 
+# OpenID Connect (OIDC) Settings
+#powerauth.nextstep.security.auth.type=OIDC
+#spring.security.oauth2.client.registration.nextstep-oidc-client.provider=nextstep-oidc-provider
```
banterCZ, line 75: Autoconfiguration has to be improved not to fail when the property is specified as empty and the type is not OIDC. In that case, the application should normally start.

powerauth-nextstep/src/main/java/io/getlime/security/powerauth/app/nextstep/configuration/SecurityConfig.java
```diff
@@ -31,19 +39,57 @@
...
      * @param http HTTP security.
      * @throws Exception Thrown when configuration fails.
      */
     @Bean
     public SecurityFilterChain filterChain(final HttpSecurity http) throws Exception {
+        if (authType == AuthType.OIDC) {
+            logger.info("Initializing OIDC authentication.");
+            http.authorizeHttpRequests(authorize -> authorize
+                            .requestMatchers(
+                                    new AntPathRequestMatcher("/login/oauth2/**"),
```
zcgandcomp, line 60: Only 

- actuator
- status 
- and the  /login/ouath2
should be permitted.


Swagger console and docs should be accessible only for authorized user (same as in PAC)

zcgandcomp (commented):
We should enable only minimum endpoint without authorization.

zcgandcomp (approved):
Looks OK.

romanstrobl (approved):

PR #1571, opened 2024-02-08
Fix #1569: NextStep Home Page

### LINKED ISSUE #1569: NextStep Home Page
NextStep is missing a home page with version, link to swagger etc.

jnpsk:
- `webapp/resources/` directory was taken from PAS.
- Details displayed on the home screen are same as response of `/api/service/status` endpoint.
- `.run/*.xml` files was modified to load static resources also for the IDE run.


banterCZ (commented):
Thanks for the home page. I am not sure whether we need so much CSS and javascript because of a simple view. @zcgandcomp Any opinion?

zcgandcomp:
> Thanks for the home page. I am not sure whether we need so much CSS and javascript because of a simple view. @zcgandcomp Any opinion?

Hello,

yes can we please minimize it? It should be static page if possible - no javascript at app.
only needed CSS.

Thank you!

jnpsk:
Removed JS files and only those css included in index.html are kept.

banterCZ (approved):
Thanks, much better without js.

jandusil (approved):
Great

zcgandcomp (approved):
Looks OK

PR #1577, opened 2024-02-13
Fix #1576: UserProfileController returns Internal Server Error if Principal missing

### LINKED ISSUE #1576: UserProfileController returns Internal Server Error if Principal missing

banterCZ:


powerauth-webflow-resources/src/main/java/io/getlime/security/powerauth/lib/webflow/resource/controller/UserProfileController.java
```diff
@@ -76,7 +74,11 @@ public UserProfileController(DataAdapterClient client, WebFlowResourcesServerCon
      * @return User profile.
      */
     @GetMapping("me")
-    public @ResponseBody UserResponse me(@AuthenticationPrincipal OAuth2AuthenticatedPrincipal principal) {
+    public UserResponse me(@AuthenticationPrincipal OAuth2AuthenticatedPrincipal principal) {
+        if (principal == null) {
+            throw new InsufficientAuthenticationException("Missing principal");
```
banterCZ, line 79: Let's return 401 instead of 500 (because of NPE).

romanstrobl (approved):
Looks OK.

PR #1616, opened 2024-04-02
Fix #1615: Remove Guava dependency

### LINKED ISSUE #1615: Remove Guava dependency
Since _owasp-java-html-sanitizer_ does not use Guava anymore, see #1614, the dependency may be removed.

A follow-up to #1535

banterCZ:


jnpsk (approved):
looks ok

PR #1637, opened 2024-04-30
Fix #1636: For API /auth/combinded assert credentialName

### LINKED ISSUE #1636: For API /auth/combinded assert credentialName
If credentialName is used in call API /auth/combined it has to be the same as the the credentialName used to generate the OTP.

we should not remove the parameter from compatibility reason, but.

- assert that the value in credentialName is the same as used for OTP generation.
- update the documentation of API - the parameter should not be used or it has to be the same as used for OTP generation.

jandusil:


romanstrobl (changes requested):
We should rather remove the parameter completely in my opinion, the parameter is ignored and should be take form the OTP.

jandusil:
@zcgandcomp As the author of the issue I would need your response/comment

zcgandcomp:
Hmm Roman is right. We can just remove it - it will still work (caller doesn't need to change the call immediately) and the functionality will not be affected - as the parameter is basically ignored. 
We can take care of notifying the callers in the migration guide. So they can remove the parameter and, if necessary, use the correct one in the creation of the OTP.

powerauth-nextstep/src/main/java/io/getlime/security/powerauth/app/nextstep/service/AuthenticationService.java
```diff
@@ -551,9 +551,6 @@ public CombinedAuthenticationResponse authenticateCombined(CombinedAuthenticatio
         final IdGeneratorService idGeneratorService = serviceCatalogue.getIdGeneratorService();
 
         final OtpEntity otp = otpService.findOtp(request.getOtpId(), request.getOperationId());
-        if (request.getCredentialName() != null && !request.getCredentialName().equals(otp.getCredentialDefinition().getName())) {
```
zcgandcomp: Can we please check that the OTP has defined credential definition and raise specific error? Otherwise we will get NPE later (it's the bug in the current implementation)

powerauth-nextstep/src/main/java/io/getlime/security/powerauth/app/nextstep/service/AuthenticationService.java
```diff
@@ -581,11 +578,9 @@ public CombinedAuthenticationResponse authenticateCombined(CombinedAuthenticatio
             operation = operationPersistenceService.getOperation(request.getOperationId());
         }
         if (request.isUpdateOperation() && operation == null) {
-            throw new InvalidRequestException("Operation not found, however operation update requested for credential and OTP: " + request.getCredentialName() + ", " + otp.getOtpDefinition().getName());
-        }
-        if (request.getOperationId() != null && !request.getOperationId().equals(otp.getOperation().getOperationId())) {
-            throw new InvalidRequestException("Operation ID mismatch for credential and OTP: " + request.getCredentialName() + ", " + otp.getOtpDefinition().getName());
+            throw new InvalidRequestException("Operation not found, however operation update requested OTP: " + otp.getOtpDefinition().getName());
```
zcgandcomp, line 580: The credential name can be obtained from the OTP object.

zcgandcomp (changes requested):
One enhancement of current possible NPE and one log message otherwise fine.

zcgandcomp (approved):
We can have it in the migration guide - the parameter is required in the OTP creation. But whoever uses it should already know.

PR #1672, opened 2024-07-31
Set Java version for GitHub Action to 8

banterCZ:


pom.xml
```diff
@@ -407,6 +405,16 @@
...
             </distributionManagement>
         </profile>
+
+        <profile>
+            <id>jdk9+</id>
+            <activation>
+                <jdk>[9,)</jdk>
+            </activation>
+            <properties>
+                <maven.compiler.release>${java.version}</maven.compiler.release>
```
banterCZ, line 415: When building by JDK 8, it would fail on this property. But when building by 9+, we need this for backward compatibility, see https://www.baeldung.com/java-compiler-release-option

pom.xml
```diff
@@ -89,9 +89,7 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <maven.compiler.source>8</maven.compiler.source>
-        <maven.compiler.target>8</maven.compiler.target>
-        <maven.compiler.release>8</maven.compiler.release>
+        <java.version>8</java.version>
```
banterCZ, line 92: Spring Boot parent defines
```xml
<maven.compiler.source>${java.version}</maven.compiler.source>
<maven.compiler.target>${java.version}</maven.compiler.target>
```

romanstrobl (approved):
Great, thanks!

PR #1684, opened 2024-08-06
Fix #1683: Provide default lang.json for external resources

### LINKED ISSUE #1683: Provide default lang.json for external resources
When I configure WebFlow following way:

```
    <Parameter name="powerauth.webflow.page.ext-resources.location" value="file:///tmp/ext-resources/"/>
```

The following error occurs in case the `lang.json` file is missing:

```
2024-08-06 14:32:31.153  WARN 40259 --- [io-8080-exec-34] i.g.s.p.app.webflow.i18n.I18NService     : Error occurred while retrieving the language setting 

java.io.FileNotFoundException: /tmp/ext-resources/lang.json (No such file or directory)
```

We should retrieve the default `lang.json` from `classpath:/static/resources/lang.json` in case it is not set externally.

romanstrobl:


powerauth-webflow/src/main/java/io/getlime/security/powerauth/app/webflow/configuration/WebMvcConfiguration.java
```diff
@@ -140,7 +141,11 @@ public ReloadableResourceBundleMessageSourceWithListing messageSource() {
 
     @Bean
     public Resource languageSettingSource() throws MalformedURLException {
-        return resourceLoader.getResource(configuration.getResourcesLocation() + "lang.json");
+        Resource resource = resourceLoader.getResource(configuration.getResourcesLocation() + "lang.json");
+        if (!resource.exists()) {
+            resource = new UrlResource("classpath:/static/resources/lang.json");
```
banterCZ, line 146: May we log this branch, please? Something like
`logger.info("lang.json not found in {}", configuration.getResourcesLocation());`

banterCZ (approved):

PR #1718, opened 2024-10-18
Fix #1717: Add support for bcrypt to store the password

### LINKED ISSUE #1717: Add support for bcrypt to store the password
Currently when we re using hashing for the protection of stored password only Argon is available. We need to support also bcrypt.

romanstrobl:


banterCZ (approved):
Looks good.

PR #1767, opened 2025-02-04
Fix #1765: NextStep: Missing securitySchemes in swagger

### LINKED ISSUE #1765: NextStep: Missing securitySchemes in swagger
### Description

Currently, the generated Swagger does not state the `securitySchemes`.

### Acceptance criteria

The generated Swagger includes `securitySchemes`, if neccessary.

### Technical specification

NextStep currently supports `NONE` or `OIDC` auth types. It would be inconvenient to propagate the security schemes via a static configuration annotation `@SecuritySchemes`. What we can do is to include the security scheme based on the configuration property `powerauth.nextstep.security.auth.type` to dynamically generate Swagger with the currently used security scheme. For example:

```java
@Bean
public OpenAPI openAPI(final ServletContext servletContext, @Value("${powerauth.nextstep.security.auth.type}") final String authType) {
        final Components components = new Components();
        if (authType.equals("OIDC")) {
            components.addSecuritySchemes("oidc", new io.swagger.v3.oas.models.security.SecurityScheme()
                    .type(io.swagger.v3.oas.models.security.SecurityScheme.Type.OPENIDCONNECT));
        }

        return new OpenAPI()
                .components(components);
}
```

### QA specification

This does not affect functionality of the NextStep, it just modifies the Swagger generation process.

jnpsk:
When the Next Step is configured with `powerauth.nextstep.security.auth.type=OIDC`, it is expected an access token is included in the request header. This PR enhances the generated swagger, when the auth type is set to OIDC as follows:

```yaml
security:
    - bearerAuth: []
    
components:
    securitySchemes:
        bearerAuth:
            type: http
            scheme: bearer
            bearerFormat: JWT
```

In case of `powerauth.nextstep.security.auth.type=NONE` (by default), no change is visible.

### DIFF
--- a/powerauth-nextstep/src/main/java/io/getlime/security/powerauth/app/nextstep/controller/ServiceController.java
+++ b/powerauth-nextstep/src/main/java/io/getlime/security/powerauth/app/nextstep/controller/ServiceController.java
@@ -24,6 +24,7 @@ import io.getlime.security.powerauth.lib.nextstep.model.response.ServiceStatusRe
 import io.swagger.v3.oas.annotations.Operation;
 import io.swagger.v3.oas.annotations.responses.ApiResponse;
 import io.swagger.v3.oas.annotations.responses.ApiResponses;
+import io.swagger.v3.oas.annotations.security.SecurityRequirements;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -72,6 +73,7 @@ public class ServiceController {
      * Controller resource with system information.
      * @return System status info.
      */
+    @SecurityRequirements
     @Operation(summary = "Get service status")
     @ApiResponses(value = {
             @ApiResponse(responseCode = "200", description = "Service status sent in response"),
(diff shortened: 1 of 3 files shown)


zcgandcomp:
Question: this will apply the security to all endpoints (in swagger). The `/api/service/status`  is always permitted, but we can probably ignore this...

powerauth-nextstep/src/main/java/io/getlime/security/powerauth/app/nextstep/controller/ServiceController.java
```diff
@@ -72,6 +73,7 @@ public void setBuildProperties(BuildProperties buildProperties) {
      * Controller resource with system information.
      * @return System status info.
      */
+    @SecurityRequirements
```
jnpsk, line 76: This annotation overrides the global security requirement settings in swagger. It adds `security: []` for the endpoint, which stands for no security.

```json
"/api/service/status": {
    "get": {
        "responses": {
            ...
        },
        "security": []
    }
}
```

jnpsk: However, after double-check, the `security: []` is present also in the case of `powerauth.nextstep.security.auth.type=NONE` for the particular endpoint only. If we would like to avoid specifying "no security" for the endpoint, when globally there is no security required, it would be necessary to solve it dynamically too.

zcgandcomp: It is not "nice", but technically correct there is no security for this endpoint. So we can use it.

romanstrobl (approved):
Looks OK.

PR #1845, opened 2025-07-14
Fix #1841: Shedlock table creation should be conditional

### LINKED ISSUE #1841: Shedlock table creation should be conditional
### Affected product(s) and version(s)

1.10.x

### Environment

_No response_

### Steps to reproduce, current result, expected result

In case the same schema is used, the addition of `shedlock` table will fail, because it is already included in PowerAuth server.

`CREATE TABLE shedlock` should probably be `CREATE TABLE shedlock IF NOT EXISTS`.

### Log output

_No response_

banterCZ:
Works for Postgres but not for Oracle


romanstrobl (approved):
Thanks for the fix.

PR #1872, opened 2025-07-28
Fix #1871: Missing Required POM Metadata

### LINKED ISSUE #1871: Missing Required POM Metadata
### Task specification

After migrating to the new Sonatype Central Portal, there is a new validation for [Required POM Metadata](https://central.sonatype.org/publish/requirements/#required-pom-metadata).

Specifically project `powerauth-mtoken-model` is missing name, which is [not inherited](https://maven.apache.org/pom.html#Inheritance) from the parent pom.

banterCZ:


jnpsk (approved):
looks good

PR #1903, opened 2025-09-30
Fix #1836: Rename io.getlime to com.wultra

### LINKED ISSUE #1836: Rename io.getlime to com.wultra
### Task specification

Rename `io.getlime` to `com.wultra`, both Java packages and Maven artifacts.

banterCZ:


romanstrobl (approved):
Looks OK.

PR #1910, opened 2025-10-04
Fix #1874: Document API changes for credentialSource and credentialTa…

### LINKED ISSUE #1874: Document API changes for credentialSource and credentialTarget
### Task specification

_No response_

zcgandcomp:
…rget

### DIFF
--- a/docs/Web-Flow-1.10.0.md
+++ b/docs/Web-Flow-1.10.0.md
@@ -15,7 +15,7 @@ Added table `shedlock` to prevent execution of the same task from another node.
 
 ### Adding Credential Source and Credential Target
 
-Credentials object holders extends of attributes `credentialSource` and `credentialTarget` with type `credentialLocation`. 
+Credentials object holders are extended with the attributes `credentialSource` and `credentialTarget` of type `credentialLocation`. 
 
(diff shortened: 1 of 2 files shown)


docs/Web-Flow-1.10.0.md
```diff
@@ -1,8 +1,40 @@
...
+This guide contains instructions for migration from PowerAuth WebFlow version `1.9.x` to version `1.10.x`.
 
 
+# 1.10.0
+
 ## Database Changes
 
 Added table `shedlock` to prevent execution of the same task from another node.
+
+# 1.10.1
```
banterCZ, line 12: 🤔 We have not yet split the migration guides by bugfix version, but why not? Just highlighting that the filename is `docs/Web-Flow-1.10.0.md`.

banterCZ (approved):
Looks good. This should be cherry-picked to develop as well.

docs/Web-Flow-1.10.0.md
zcgandcomp: Yes, but as we are going to do several patchsets, we should start using the migration in an incremental way..

PR #1937, opened 2025-10-21
Fix #1936 Automatic publish in Sonatype

### LINKED ISSUE #1936: Automatic publish in Sonatype
### Task specification

_No response_

filipbursik-wultra:


powerauth-nextstep/pom.xml
```diff
@@ -239,10 +239,12 @@
             <build>
                 <plugins>
                     <plugin>
-                        <groupId>org.apache.maven.plugins</groupId>
-                        <artifactId>maven-deploy-plugin</artifactId>
+                        <groupId>org.sonatype.central</groupId>
+                        <artifactId>central-publishing-maven-plugin</artifactId>
+                        <version>0.9.0</version>
```
banterCZ: I would rather see the version in the properties and in the plugin management.

powerauth-webflow-client/pom.xml
```diff
@@ -164,10 +164,12 @@
             <build>
                 <plugins>
                     <plugin>
-                        <groupId>org.apache.maven.plugins</groupId>
-                        <artifactId>maven-deploy-plugin</artifactId>
+                        <groupId>org.sonatype.central</groupId>
```
banterCZ, line 167: So this plugin overrides the phase `deploy`, or are we going to call it differently?

powerauth-nextstep/pom.xml
```diff
@@ -239,10 +239,12 @@
...
+                        <extensions>true</extensions>
```
banterCZ, line 244: Interesting. Do you have any details on why this is required?

powerauth-nextstep/pom.xml
filipbursik-wultra: It should be modified.

powerauth-webflow-client/pom.xml
filipbursik-wultra: This overrides the `deploy` phase, so there’s no need to call it differently.

powerauth-nextstep/pom.xml
filipbursik-wultra: By default, it’s uploaded to Sonatype only in the staging phase, and (probably for security reasons, to avoid releasing something unintentionally) you need to go into the UI to publish it. However, according to the Sonatype documentation – https://central.sonatype.org/publish/publish-portal-maven/
 – this can be handled using the plugin.

banterCZ (approved):
Looks good.

korbelm (approved):
Looks OK

PR #1948, opened 2025-10-29
Fix #1947: Fix LDAP healthcheck issue

### LINKED ISSUE #1947: Fix LDAP healthcheck issue
### Task specification

Disable LDAP healthcheck from the auto healthcheck configuration.

zcgandcomp:


banterCZ (approved):
If we are in a hurry, we may merge. But I will try to figure out a way to enable health check only if LDAP is configured.

PR #1953, opened 2025-11-05
Fix #568: Fix maven central publish

### LINKED ISSUE #568: Validate formData when creating new operations
Next Step should not allow creating new operations with `null` value of `formData`. At least the mandatory parameters should be validated.

filipbursik-wultra:
The timeout for publishing has been increased to 5 hours, and the skip publishing option is now applied only to a specific profile.

Depends on https://github.com/wultra/wultra-infrastructure/pull/64

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -497,6 +497,7 @@
                             <publishingServerId>${env.server_id}</publishingServerId>
                             <autoPublish>true</autoPublish>
                             <waitUntil>published</waitUntil>
+                            <waitMaxTime>${env.wait_max_time}</waitMaxTime>
                         </configuration>
                     </plugin>
                 </plugins>
--- a/powerauth-webflow-authentication-consent/pom.xml
+++ b/powerauth-webflow-authentication-consent/pom.xml
@@ -51,17 +51,9 @@
         </dependency>
     </dependencies>
 
-    <build>
-        <plugins>
-            <plugin>
-                <groupId>org.sonatype.central</groupId>
-                <artifactId>central-publishing-maven-plugin</artifactId>
-                <extensions>true</extensions>
-                <configuration>
-                    <skipPublishing>true</skipPublishing>
-                </configuration>
-            </plugin>
-        </plugins>
-    </build>
+    <properties>
+        <skipPublishing>true</skipPublishing>
+        <maven.deploy.skip>true</maven.deploy.skip>
+    </properties>
 
 </project>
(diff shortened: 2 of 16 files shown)


banterCZ (approved):
Looks good.

korbelm (approved):
great, thanks

zcgandcomp (approved):
If the public-repository profile is activated, set the snatype plugin to skip the deploy.
For ALL modules, that should not be deployed.

powerauth-webflow-authentication-sms/pom.xml
```diff
@@ -47,14 +47,38 @@
         </dependency>
     </dependencies>
 
+    <profiles>
+        <profile>
+            <id>public-repository</id>
```
banterCZ: 🤔 Isn't it possible to define it only once at the parent pom?

filipbursik: The profile and plugin are also defined in the parent POM, so the plugin needs to be disabled in the POM files of specific projects.

powerauth-webflow-authentication-consent/pom.xml
```diff
@@ -51,17 +51,9 @@
...
-                <extensions>true</extensions>
-                <configuration>
-                    <skipPublishing>true</skipPublishing>
-                </configuration>
-            </plugin>
-        </plugins>
-    </build>
+    <properties>
+        <skipPublishing>true</skipPublishing>
+        <skip>true</skip>
```
banterCZ: Shouldn't this be `<maven.deploy.skip>true</maven.deploy.skip>`? See https://maven.apache.org/plugins/maven-deploy-plugin/deploy-mojo.html

filipbursik-wultra: Fixed

pom.xml
```diff
@@ -497,6 +497,7 @@
                             <publishingServerId>${env.server_id}</publishingServerId>
                             <autoPublish>true</autoPublish>
                             <waitUntil>published</waitUntil>
+                            <waitMaxTime>${env.wait_max_time}</waitMaxTime>
```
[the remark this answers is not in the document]
banterCZ: 🤔 I think that Copilot is wrong, because `skipPublishing` is a property defined by `central-publishing-maven-plugin`, see https://mavenplugins.github.io/central-publishing-maven-plugin/publish-mojo.html

zcgandcomp (approved):
The copilot has a relevant point. We can disable in the profile all deploy plugins. But we do not need to. The current version is fine.

korbelm (approved):

PR #1998, opened 2026-02-08
Issue #1996: Change base image to IBM Semeru and add MSSQL JDBC driver

filipbursik-wultra:


powerauth-nextstep/docker/Dockerfile
```diff
@@ -1,17 +1,20 @@
-FROM eclipse-temurin:21-jre
```
filipbursik-wultra, line 1: @banterCZ @korbelm Why does nextstep use a different base image with a different java runtime? 🤔

korbelm: it should be just "openjdk" base, this was copy pasted from powerauth-docker where we used standalone tomcats, but now we want embeded  pls use same as for pac and others

filipbursik-wultra: Yes, the base image has been updated to `ibm-semeru-runtimes:open-21.0.9_10-jre-noble`.

korbelm (approved):

PR #2080, opened 2026-05-19
Fix #2079: Reset Credentials should follow the Credentials Source logic

### LINKED ISSUE #2079: Reset Credentials should follow the Credentials Source logic
### Description

As of now, when reset credentials are invoked, the Credentials are reset, but the Credential Source is not honored.

If the password is reset and the CredentialSource is `PROXY` and the CredentialTarget is `LOCAL`  the source should change. If the target is PROXY the NextStep should log warn

### Acceptance criteria

- The credential source should be changed if the target is LOCAL.
- The feature is documented in the Migration Guide
- The feature is documented in the API description

### Technical specification

_No response_

### QA specification

_No response_

zcgandcomp:


powerauth-nextstep/src/main/java/com/wultra/security/powerauth/app/nextstep/service/CredentialService.java
```diff
@@ -446,6 +446,16 @@ public ResetCredentialResponse resetCredential(ResetCredentialRequest request) t
...
         }
+        // only target local credential can be reset
+        if (CredentialLocation.LOCAL == credential.getTarget()) {
+            // if source was proxy change the information about the source
+            if (CredentialLocation.LOCAL != credential.getSource()) {
+                credential.setSource(CredentialLocation.LOCAL);
+            }
+        } else {
+            logger.warn("invalid attempt to set credential value for external password, userId: {}", user.getUserId());
+        }
```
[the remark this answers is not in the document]
zcgandcomp: This is intentional - the integrator can reset credentials and update the source manually. The exception was not used to prevent impact on the current integration. But a valid comment indeed.

powerauth-nextstep/src/main/java/com/wultra/security/powerauth/app/nextstep/service/CredentialService.java
```diff
@@ -246,7 +246,7 @@ public UpdateCredentialResponse updateCredential(UpdateCredentialRequest request
                 credential.setTimestampLastCredentialChange(changeTimestamp);
                 updateCredentialExpiration = true;
             } else {
-                logger.warn("invalid attempt to set credential value for external password, userId: {}", user.getUserId());
+                logger.warn("Attempt to update external credentials, credentials updated, external credential value is NOT updated, userId: {}", user.getUserId());
```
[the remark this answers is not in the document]
zcgandcomp: Resoned to the current version of message.

powerauth-nextstep/src/main/java/com/wultra/security/powerauth/app/nextstep/service/CredentialService.java
```diff
@@ -446,6 +446,15 @@ public ResetCredentialResponse resetCredential(ResetCredentialRequest request) t
         if (request.getCredentialType() != null) {
             credential.setType(request.getCredentialType());
         }
+        if (CredentialLocation.LOCAL == credential.getTarget()) {
+            // if source was not local change the information about the source
+            if (CredentialLocation.LOCAL != credential.getSource()) {
+                credential.setSource(CredentialLocation.LOCAL);
+            }
+        } else {
+            logger.warn("Attempt to reset external credentials, credentials reset, external credential value is NOT updated, userId: {}", user.getUserId());
```
pavsindelar, line 455: I don't see how the rest of the logic is influenced by the fact that the credential target is not local.

Shouldn't the credential value remain the same in case `CredentialLocation.LOCAL != credential.getTarget()`?

zcgandcomp: We are still maintaining the whole credential record - including the credential value. This should be updated.
But we are not able to reset credentials in external system.
AND most importantly (the purpose of this ticket) - when we reset the credentials with target LOCAL, we should properly update the credential source. This is similar to change password, if password is changed with target LOCAL the source is set LOCAL.

We are not throwing an exception (also an option) as we want to reset the status of credentials, etc.
We are resetting the local credential value - in case somebody switches the source manually.

pavsindelar: Ok, thanks for explanation. I guess reseting the local credentials does no harm even if the target is external 👍  And the possibility to switch to the local credentials manually provides business justification for resetting the value.

pavsindelar (approved):
Looks good 👍

